PR #33, opened 2025-10-25
War Supply Farm + Gemstone Farm improvements + code updates

Gahais:
hi :)
Thank you for creating this useful hub for Guild Wars bots

I rewrote War Supply farm to work with your BotsHub and added some improvements to Gemstone farm and added some more improvements to general code
If you don't like this code more or less then you can tell me what to refactor, of course, but some code changes might be interesting, I think
If you have some uncommitted changes then you can commit them first and I can reapply merge on top of your changes, there is no hurry with this stuff
I can help fix bugs or write some more farm bots, I can try rewriting these bots to farm single gemstones to be working with your BotsHub

These are my humble suggestions, the choice for each one is up to you, of course
- changed remaining null occurrences into Null for consistency, SciTe probably also suggests Null like True, False, I think
- changed RndSleep into RandomSleep for clarity, Rnd might be a bit ambiguous with words like rendering, rounding, imho
- changed bag_number into bags_count, same string length, bag_number feels a bit like index of some specific numbered bag, count feels more like quantity, imho
- changed group into party for all occurrences that refer to team for disambiguation. I think party can be good synonym of team because P button open party menu in game. And groups might be used to describe groups of foe mobs
- changed PlayerNumber into ModelID for agent structure, because I think that most of the time this attribute is used for detecting npc/item agents different than human players
- changed inventory management to happen before farm runs rather than after to avoid going into runs with full equipment and to use convenient inventory management even if run fails

Below some more changes:
- Added $SUCCESS, $FAIL, $PAUSE, NOT_STARTED constants for clarity to easily manage return states from farms and indicate fails or successes in other functions
- Added convenient travelling and returning with resigning functions with map for location ids and location names
- Added some disconnection handlers just in case for moving/aggroing loops. Didn't test it out, though, it's optional, I can remove these handlers
- Added max timers for runs in Gemstone and War Supply farms. Interesting idea if such max timers can be used globally, but that would need some way to interrupt farm from continuing, maybe could use some background for this
- Changed indexing for all 1 dimensional arrays in GWA2 and elsewhere to start from 0 instead of 1 and removed these array size checks using first element of array. Imho, this might be safer in case when some other programmers add some more farms into this BotsHub in pull requests and may not know that in AutoIT these arrays are often indexed from 1 instead of 0 and that first element in AutoIT array is considered a count of elements in array (like these arrays returned from StringSplit). This can be confusing for programmers with background in other languages. Maybe later I can also refactor these 2D arrays returned from ProcessList, WinList to remove first element as counter from them. UBound() can be used alternatively. I could write some custom functions for managing 2D arrays like indexed from 0
- Changed usages of maps into usages of dictionary maps. The choice is yours which should be used. I think dictionaries can be useful because they are automatically passed as object references without copying elements inside dictionary map. So this probably should speed up most often used functions like these for moving and aggroing, so that they don't copy anything on each call but pass the same (or cloned once) dictionary map. Also dictionary map object can be Null so it might be easy to set default Null optional parameter in function definitions
- Changed fighting functions a bit, I think, there were 3 similar fight functions implementations, 2 in Utils and 1 in Gemstone farm, so I merged them all into 1 generic function that can have any parameters passed into it as 1 optional dictionary map. These parameters can be easily extended
- automatic GUI positioning in the middle of the screen. I read that setting -1, -1 to x, y automatically positions GUI in the middle of the screen in the AutoIT
- Added to the GUI success counter to ensure that fails + successes = number of runs, if less than number of runs then this can help point out to some bugged situation
- run duration is printed into GUI console for convenience
- Added counters for all items, that might be possibly found during farm runs, including festival items (the list might be extended in the future). Counters also for player's gold and gold items, that are calculated by adding delta income to total sum after each run

Below should be a screenshot of how the GUI looks like now, with success counter and with all items counters

I see that JackLinesMatthews creates some interesting fork of your Botshub project in which he added some more functionalities to looting and salvaging + minotaur horn farm

screen_war_supply

### DIFF
--- a/lib/Utils.au3
+++ b/lib/Utils.au3
@@ -192,7 +192,7 @@ EndFunc
 
 #Region Loot items
 ;~ Loot items around character
-Func PickUpItems($DefendFunction = Null, $ShouldPickItem = DefaultShouldPickItem, $range = $RANGE_COMPASS)
+Func PickUpItems($defendFunction = Null, $shouldPickItem = DefaultShouldPickItem, $range = $RANGE_COMPASS)
 	If (GUICtrlRead($GUI_Checkbox_LootNothing) == $GUI_CHECKED) Then Return
 
 	Local $item
@@ -207,8 +207,8 @@ Func PickUpItems($DefendFunction = Null, $ShouldPickItem = DefaultShouldPickItem
 		$agentID = DllStructGetData($agent, 'ID')
 		$item = GetItemByAgentID($agentID)
 
-		If ($ShouldPickItem($item)) Then
-			If $DefendFunction <> Null Then $DefendFunction()
+		If ($shouldPickItem($item)) Then
+			If $defendFunction <> Null Then $defendFunction()
 			If Not GetAgentExists($agentID) Then ContinueLoop
 			PickUpItem($item)
 			$deadlock = TimerInit()
@@ -332,7 +332,7 @@ EndFunc
 
 
 ;~ Find and open chests in the given range (earshot by default)
-Func FindAndOpenChests($range = $RANGE_EARSHOT, $DefendFunction = Null, $BlockedFunction = Null)
+Func FindAndOpenChests($range = $RANGE_EARSHOT, $defendFunction = Null, $blockedFunction = Null)
 	If IsPlayerDead() Then Return
 	If FindInInventory($ID_Lockpick)[0] == 0 Then Return
 	Local $gadgetID
@@ -347,7 +347,7 @@ Func FindAndOpenChests($range = $RANGE_EARSHOT, $DefendFunction = Null, $Blocked
 			;MoveTo(DllStructGetData($agent, 'X'), DllStructGetData($agent, 'Y'))		;Fail half the time
 			;GoSignpost($agent)															;Seems to work but serious rubberbanding
 			;GoToSignpost($agent)															;Much better solution BUT character doesn't defend itself while going to chest + function kind of sucks
-			GoToSignpostWhileDefending($agent, $DefendFunction, $BlockedFunction)			;Final solution
+			GoToSignpostWhileDefending($agent, $defendFunction, $blockedFunction)			;Final solution
 			If IsPlayerDead() Then Return
 			RandomSleep(200)
 			OpenChest()
@@ -381,7 +381,7 @@ EndFunc
 
 
 ;~ Go to signpost and wait until you reach it.
-Func GoToSignpostWhileDefending($signpost, $DefendFunction = Null, $BlockedFunction = Null)
+Func GoToSignpostWhileDefending($signpost, $defendFunction = Null, $blockedFunction = Null)
 	Local $me = GetMyAgent()
 	Local $X = DllStructGetData($signpost, 'X')
 	Local $Y = DllStructGetData($signpost, 'Y')
@@ -389,11 +389,11 @@ Func GoToSignpostWhileDefending($signpost, $DefendFunction = Null, $BlockedFunct
 	While IsPlayerAlive() And GetDistance($me, $signpost) > 250 And $blocked < 15
 		Move($X, $Y, 100)
 		RandomSleep(GetPing() + 50)
-		If $DefendFunction <> Null Then $DefendFunction()
+		If $defendFunction <> Null Then $defendFunction()
 		$me = GetMyAgent()
 		If Not IsPlayerMoving() Then
-			If $BlockedFunction <> Null And $blocked > 10 Then
-				$BlockedFunction()
+			If $blockedFunction <> Null And $blocked > 10 Then
+				$blockedFunction()
 			EndIf
 			$blocked += 1
 			Move($X, $Y, 100)
@@ -2350,7 +2350,7 @@ Func MoveAggroAndKillGroups($foes, $firstGroup, $lastGroup)
 	If IsPlayerAndPartyWiped() Then Return $FAIL
 	If Not IsArray($foes) Or UBound($foes, $UBOUND_DIMENSIONS) <> 2 Then Return $FAIL
 	If UBound($foes, $UBOUND_COLUMNS) <> 3 And UBound($foes, $UBOUND_COLUMNS) <> 4 Then Return $FAIL
-	If $firstGroup < 1 Or UBound($foes) < $lastGroup  Then Return $FAIL
+	If $firstGroup < 1 Or UBound($foes) < $lastGroup Then Return $FAIL
 	If $firstGroup > $lastGroup Then Return $FAIL
 	Local $x, $y, $log, $range
 	For $i = $firstGroup - 1 To $lastGroup - 1 ; Caution, groups are indexed from 1, but $foes array is indexed from 0


src/Farm-EdenIris.au3
```diff
@@ -33,88 +35,71 @@ Global $IRIS_FARM_SETUP = False
...
 	Local $agents = GetAgentArray(0x400)
-	For $i = 1 To $agents[0]
-		Local $agent = $agents[$i]
-		Local $agentID = DllStructGetData($agent, 'ID')
+	For $i = 0 To UBound($agents) - 1
+		Local $agentID = DllStructGetData($agents[$i], 'ID')
 		$item = GetItemByAgentID($agentID)
 		If (DllStructGetData($item, 'ModelID') == $ID_Red_Iris_Flower) Then
-			Info('Iris: (' & Round(DllStructGetData($agent, 'X')) & ',' & Round(DllStructGetData($agent, 'Y')) & ')')
+			Info('Iris: (' & Round(DllStructGetData($agents[$i], 'X')) & ',' & Round(DllStructGetData($agents[$i], 'Y')) & ')')
```
caustic-kronos: Might be better to keep a variable $agent to keep code a bit lighter and more readable.

src/Farm-Vanguard.au3
```diff
@@ -23,183 +25,186 @@
...
+	If MoveAggroAndKill(-12373, 12899, 'Move To Start') == $FAIL Or _
+		MoveAggroAndKill(-9464, 15937, 'Charr Group') == $FAIL Or _
+		MoveAggroAndKill(-9130, 13535, 'Moving') == $FAIL Or _
+		MoveAggroAndKill(-5532, 11281, 'Moving') == $FAIL Or _
+		MoveAggroAndKill(-3979, 9184, 'Mantid Group') == $FAIL Or _
+		MoveAggroAndKill(-355, 9296, 'Again mantid Group') == $FAIL Or _
+		MoveAggroAndKill(836, 12171, 'Charr Patrol') == $FAIL Or _
+		MoveAggroAndKill(884, 15641, 'Charr Group') == $FAIL Or _
+		MoveAggroAndKill(2956, 10496, 'Mantid Group') == $FAIL Or _
+		MoveAggroAndKill(5160, 11032, 'Moving') == $FAIL Then Return $FAIL
```
caustic-kronos: There are lots of places with tons of MoveAggroAndKill following each other, I think rather than put them all into a big If I would rather make big arrays and loop to call MoveAggroAndKill on them

src/Farm-WarSupplyKeiran.au3
```diff
@@ -0,0 +1,434 @@
+#CS = Comment Block Start ==========================================================================
+; BotsHub Author: caustic-kronos (aka Kronos, Night, Svarog)
```
caustic-kronos: You can skip BotsHub Author, each file has its own Author and this one's entirely yours

lib/Utils.au3
```diff
@@ -61,10 +64,10 @@ Func RunTests($STATUS)
 	;Info(DllStructGetData(GetEffect($ID_Shroud_of_Distress), 'TimeStamp'))
 	;Info(GetEffectTimeRemaining(GetEffect($ID_Shroud_of_Distress)))
 	;Info(_dlldisplay(GetEffect($ID_Shroud_of_Distress)))
-	;RndSleep(1000)
+	;RandomSleep(1000)
 
-	;Return 0
-	Return 2
+	;Return $SUCCES
```
caustic-kronos:
```suggestion
	;Return $SUCCESS
```

lib/Utils.au3
```diff
@@ -1679,287 +1767,361 @@ EndFunc
...
 
-		_ArrayAdd($returnAgents, $curAgent)
-		$returnAgents[0] += 1
+	If $coordX == Null Or $coordY == Null Then
+		Local $me = GetMyAgent()
+		$coordX = DllStructGetData($me, 'X')
+		$coordY = DllStructGetData($me, 'Y')
+	EndIf
+	For $i = 0 To UBound($agents) - 1
+		If $npcAllegiance <> Null And DllStructGetData($agents[$i], 'Allegiance') <> $npcAllegiance Then ContinueLoop
```
caustic-kronos: Same reasoning here, creating a variable $agent is a small price for the clarity provided in the next lines

lib/Utils.au3
```diff
@@ -2047,317 +2202,519 @@ Func MoveAvoidingBodyBlock($coordX, $coordY, $timeOut)
...
+$Default_MoveAggroAndKill_Options.Add('flagHeroesOnFight', False)
+$Default_MoveAggroAndKill_Options.Add('callTarget', True)
+$Default_MoveAggroAndKill_Options.Add('priorityMobs', False)
+$Default_MoveAggroAndKill_Options.Add('skillsMask', Null)
+$Default_MoveAggroAndKill_Options.Add('skillsCostMap', Null)
+$Default_MoveAggroAndKill_Options.Add('skillsCastTimeMap', Null)
+$Default_MoveAggroAndKill_Options.Add('lootInFights', False)
+$Default_MoveAggroAndKill_Options.Add('openChests', True)
+$Default_MoveAggroAndKill_Options.Add('chestOpenRange', $RANGE_SPIRIT)
+$Default_MoveAggroAndKill_Options.Add('fightDuration', 60000) ;~ default 60 seconds fight duration
```
caustic-kronos: I'm a big map fan and I'm not that familiar with dictionaries, do dictionaries have an advantage over maps ?
In any case, great use of this map/dictionary, that's exactly what I had in mind when making it.

lib/GWA2.au3
```diff
@@ -3308,45 +3347,37 @@ Func GetNearestAgentToCoords($X, $Y, $agentType = 0, $agentFilter = Null)
...
-		$result[$index] = $agentArray[$i]
-		$index += 1
+Func GetParty($agents = Null)
+	Local $resultArray[0] ;~ dynamic 1D array of agents, indexed from 0
+	If $agents == Null Then $agents = GetAgentArray(0xDB)
+
+	For $i = 0 To UBound($agents) - 1
+		If DllStructGetData($agents[$i], 'Allegiance') <> 1 Then ContinueLoop
+		If Not BitAND(DllStructGetData($agents[$i], 'TypeMap'), 0x20000) Then ContinueLoop
+		_ArrayAdd($resultArray, $agents[$i]) ;~ addition to dynamic array with automatic resizing
```
caustic-kronos: Same reasoning concerning _ArrayAdd, it would be preferable to avoid it.

caustic-kronos:
Also, to answer you :
> changed remaining null occurrences into Null for consistency, SciTe probably also suggests Null like True, False, I think

✔️ 
> changed RndSleep into RandomSleep for clarity, Rnd might be a bit ambiguous with words like rendering, rounding, imho

✔️
> changed bag_number into bags_count, same string length, bag_number feels a bit like index of some specific numbered bag, count feels more like quantity, imho

✔️ 
> changed group into party for all occurrences that refer to team for disambiguation. I think party can be good synonym of team because P button open party menu in game. And groups might be used to describe groups of foe mobs

✔️
> changed PlayerNumber into ModelID for agent structure, because I think that most of the time this attribute is used for detecting npc/item agents different than human players

✔️
> changed inventory management to happen before farm runs rather than after to avoid going into runs with full equipment and to use convenient inventory management even if run fails

✔️
> Added $SUCCESS, $FAIL, $PAUSE, NOT_STARTED constants for clarity to easily manage return states from farms and indicate fails or successes in other functions

✔️
> Added convenient travelling and returning with resigning functions with map for location ids and location names

✔️ 
> Added some disconnection handlers just in case for moving/aggroing loops. Didn't test it out, though, it's optional, I can remove these handlers

➡️ I'm not sure, didn't see that, I will check and come back to you for this.
> Added max timers for runs in Gemstone and War Supply farms. Interesting idea if such max timers can be used globally, but that would need some way to interrupt farm from continuing, maybe could use some background for this

✔️ Yeah it would be nice to be able to interrupt farms (whether on a timer or from button press) but AutoIt being single threaded, it would require bots to do frequent checks by themselves. This could be handled by a AdlibRegister but I noticed when tinkering with those that too many of them cause heavy issues, so the less we use them the better.
> Changed indexing for all 1 dimensional arrays in GWA2 and elsewhere to start from 0 instead of 1 and removed these array size checks using first element of array. Imho, this might be safer in case when some other programmers add some more farms into this BotsHub in pull requests and may not know that in AutoIT these arrays are often indexed from 1 instead of 0 and that first element in AutoIT array is considered a count of elements in array (like these arrays returned from StringSplit). This can be confusing for programmers with background in other languages. Maybe later I can also refactor these 2D arrays returned from ProcessList, WinList to remove first element as counter from them. UBound() can be used alternatively. I could write some custom functions for managing 2D arrays like indexed from 0

✔️  Completely agreed, I really don't like this behaviour in AutoIt.
> Changed usages of maps into usages of dictionary maps. The choice is yours which should be used. I think dictionaries can be useful because they are automatically passed as object references without copying elements inside dictionary map. So this probably should speed up most often used functions like these for moving and aggroing, so that they don't copy anything on each call but pass the same (or cloned once) dictionary map. Also dictionary map object can be Null so it might be easy to set default Null optional parameter in function definitions

➡️ I'm not sure. I think maps can be passed as references too, or even be static or a global variables.
> Changed fighting functions a bit, I think, there were 3 similar fight functions implementations, 2 in Utils and 1 in Gemstone farm, so I merged them all into 1 generic function that can have any parameters passed into it as 1 optional dictionary map. These parameters can be easily extended

✔️ Neat !
> automatic GUI positioning in the middle of the screen. I read that setting -1, -1 to x, y automatically positions GUI in the middle of the screen in the AutoIT

✔️ 
> Added to the GUI success counter to ensure that fails + successes = number of runs, if less than number of runs then this can help point out to some bugged situation

✔️
> run duration is printed into GUI console for convenience

✔️
> Added counters for all items, that might be possibly found during farm runs, including festival items (the list might be extended in the future). Counters also for player's gold and gold items, that are calculated by adding delta income to total sum after each run. Below should be a screenshot of how the GUI looks like now, with success counter and with all items counters

✔️ 
> I see that JackLinesMatthews creates some interesting fork of your Botshub project in which he added some more functionalities to looting and salvaging + minotaur horn farm

➡️  Amazing ! I will need to take a look at it.

Gahais:
hi, thanks for feedback :)
I like your idea of collecting all existing guild wars bots into one big hub project maintained on github instead of dealing with zip files, and appreciate your refactor of this old code of GWA2, etc.
I like your code style. I'm also in favor of longer, more descriptive names. I too have some background from java, but probably I'm more C++ dude :)

hmm, do You prefer if I do interactive rebase and edit these old commits to reflect your new suggestions and say here what I changed or to not edit these old commits and make on top of these old commits new small commits, which should show what is changed?
interactive rebasing with force pushing may help keep more clean commits history but it might be harder to see what was recently changed


>     * Why the addition of :
> ```
> #CS = Comment Block Start ==========================================================================
> #CE = Comment Block End ============================================================================
> ```

This is purely for aesthetics. I can remove that if you want. I like long, descriptive names, and try to avoid short abbreviations. Like in Java where there is ArrayIndexOutOfBoundsException instead of Aiobe
Can be hard to understand from first glance what CS and CE is, and it seems it is Comment Start and Comment End, so wrote it for explanation. Can refactor that back


>     * Why adding ~ to comments on lines ? Usually, it's present on function comments to give a description, but on lines does it serve a purpose ?

sorry, I didn't notice that ;~ was present only for function comments and ; was meant for other comments. It didn't server a purpose. I just tried to use ;~ everywhere
Ok, I can refactor it back. Do You prefer convention where ;~ is for function description and ; everywhere else or second convention in which ;~ is used everywhere. 
Perhaps this original convention is better and my new convention is too crude and hard to maintain


>     * Why create Attributes_Array ?

I added Attributes_Array mostly for convenience. Also with this array can now just do that UBound(Attributes_Array) to know how much attributes there is in guild wars. So this way can avoid to use one less magical constant 44 - number of attributes. Though, I forgot to change occurrences of 	`For $i = 0 To 44`
to 	`For $i = 0 To UBound(Attributes_Array)`


> * Why IsIdentified(Rarity)Item ?

I also added that for convenience. Can remove this if you want. 
I had a problem that somehow currently bot doesn't sell identified gold items. Probably It is possible to do now these 2 options. Either store all unidentified items in chest or either keep all gold identified items in inventory. Maybe I overlooked something. So I added these improvised functions to Utils-Storage-Bot.au3 code
And I also have these 3 lines in ActiveInventoryManagement() function, only 3 lines from this file I didn't commit

```
Func ActiveInventoryManagement()
	.....
		SellItemsToMerchant() ; already in the code		
		SellItemsToMerchant(IsIdentifiedGoldItem) ; temporary line
		SellItemsToMerchant(IsIdentifiedBlueItem) ; temporary line
		SellItemsToMerchant(IsIdentifiedPurpleItem) ; temporary line 
```

I see that you and JackLinesMatthews are already working on some inventory management and looting code, so I didn't want to mess with your new code with my own further inventory management ideas. So I do other parts of BotsHub project waiting out in case you commit new inventory management code. There is no rush with this of course.

 
> Don't forget to add yourself to the CREDITS.md file, and there are probably some other files where you can credit yourself too, don't hesitate to do that.

ok, thanks, I will try to do that :)


> ✔️ It makes sense. I don't know about SciTe, I use Visual Studio Code, but it's logical to harmonize those values syntax.

cool :) is Visual Studio good? I probably never used it
I'm currently using free CLion community edition for open source like this. There is free CLion quite recently since may 2025. 
SciTe is quite hard to use imho

> ✔️ Sure. If you want to run just the inventory management part of the bot there is the Storage bot you can run.

ok, I will test out this Storage bot. Didn't try it yet

> > Added some disconnection handlers just in case for moving/aggroing loops. Didn't test it out, though, it's optional, I can remove these handlers
> 
> ➡️ I'm not sure, didn't see that, I will check and come back to you for this.

Hey, yeah, this one is quite interesting topic. I tested it out recently. So it is currently probably a bad code to add, unless we manage to fix GetMapLoading() function. Because I tested it and GetMapLoading() always returns 0. So I suspect that GetMapLoading() function is still broken after that guild wars update in 2024 probably. Fixing GetMapLoading() would be useful for gemstone farms because all 4 DoA areas have the same map ID like Gate of Anguish 474, so hard to say if player is currently in one of 4 DoA explorable areas or in outpost and whether should travel. Made currently a workaround that X,Y coordinates of player are compared with start coordinates of these 4 DoA areas. Will commit that here soon
In old DoA gemstones zip project I found this line
`Global Enum $INSTANCETYPE_OUTPOST, $INSTANCETYPE_EXPLORABLE, $INSTANCETYPE_LOADING`
so $INSTANCETYPE_OUTPOST is meant to be 0, $INSTANCETYPE_EXPLORABLE meant to be 1, $INSTANCETYPE_LOADING meant to be 2
So currently this code this code does nothing, and disconnect handler isn't triggered, GetMapLoading() says map is Outpost == 0 even if map is truly loading

I noticed that GetTarget() function is similarly broken as GetMapLoading(). GetTarget() always returns 0 even when player or foe targets some other agent

So the big question is this. Can these GetMapLoading() and GetTarget() functions be fixed?
I don't quite know yet all this low-level, assembly stuff. For example, how guys like MrJambix, gigi fix these headers in low-level byte code after there is guild wars update? How to fix/prevent guild wars crashes? Like this one
```
*--> Crash <--*
Exception: c0000005 
Memory at address fffffff0 could not be written
App: Gw.exe
BaseAddr: 004E0000
ProgramId: 1
Build: 37628
```


> ✔️ Yeah it would be nice to be able to interrupt farms (whether on a timer or from button press) but AutoIt being single threaded, it would require bots to do frequent checks by themselves. This could be handled by a AdlibRegister but I noticed when tinkering with those that too many of them cause heavy issues, so the less we use them the better.

yeah, I also noticed some weird issues when using AdlibRegister and AdlibUnregister. Since then I'm afraid to use
I also noticed in old DoA zip project that global timer is used in such "background" task to show real elapsed time of bot, every second. Is this the bad from the same reason? So maybe let's not use such global, accurate timer to avoid using AdlibRegister and AdlibUnregister?
This is this probably fragile code from old DoA zip project
```
Main()
While 1
	  AdlibRegister("TimeUpdater", 1000)
.....
Func TimeUpdater()
	$Seconds += 1
	.....
	GUICtrlSetData($TotalTimeCount, $L_Hour & ":" & $L_Min & ":" & $L_Sec)
EndFunc
```

 
> > Changed indexing for all 1 dimensional arrays in GWA2 and elsewhere to start from 0 instead of 1 and removed these array size checks using first element of array. 
> .....
> ✔️ Completely agreed, I really don't like this behaviour in AutoIt.

Yeah, I don't also like that behaviour of AutoIT. Frankly, it is the first time in my life I see something like that. ok, so maybe in next commits, I will try to also refactor these 2D arrays returned from ProcessList, WinList to remove first element as counter from them. And maybe write some custom functions for managing 2D arrays like indexed from 0. And you could revise that when I finish :)


> > Changed usages of maps into usages of dictionary maps. The choice is yours which should be used. I think dictionaries can be useful because they are automatically passed as object references without copying elements inside dictionary map. So this probably should speed up most often used functions like these for moving and aggroing, so that they don't copy anything on each call but pass the same (or cloned once) dictionary map. Also dictionary map object can be Null so it might be easy to set default Null optional parameter in function definitions
> 
> ➡️ I'm not sure. I think maps can be passed as references too, or even be static or a global variables.

hmm, sorry, didn't consider this regarding static, global variables 
I can refactor the code back to original usage of maps. Mostly I have problem with some limitation of AutoIT which I couldn't overcome. 
In AutoIT I noticed some problem that I couldn't do 2 things simultaneously. Could either do first or second but not both simultaneously. Namely, to give possibility to pass Null "optional" map to the function like below, something likestd:: optional in C++
And secondly to pass map as reference so that its elements aren't copied over and over. Especially in this new, "smart" implementation, with variants of MoveAggroAndKillInRange getting map and passing map to MoveAggroAndKill which then pass this map to fightFunction(), quite often in each loop iteration. I wanted to avoid copying of map elements over and over for optimization, to increase performance. I didn't test out if passing maps by reference actually increases performance, correct me if I'm wrong/
In AUtoIT there is ByRef keyword that is probably supposed to pass arguments as reference without copying
Could do with maps (ByRef $options), or ($options = Null) but not both (ByRef $options = Null)
Func MoveAggroAndKill($x, $y, $log = '', ByRef $options = Null) ;this probably would give an error in AutoIT when using maps, but probably not when using dicrionary because dictionary is by default an object reference and ByRef doesn't need to be used at all
So, you can tell me what you think about it and I can refactor that


> > Added to the GUI success counter to ensure that fails + successes = number of runs, if less than number of runs then this can help point out to some bugged situation
> ✔️ Sure, was there times when fails + successes was different from the total number of runs though ?

Honestly, I never noticed that fails + successes was different from the total number of runs


> > I see that JackLinesMatthews creates some interesting fork of your Botshub project in which he added some more functionalities to looting and salvaging + minotaur horn farm
> ➡️ Amazing ! I will need to take a look at it.

ok, maybe JackLinesMatthews will do some pull request of his code after we handle this pull request :)
I saw some interesting loot, salvage improvements in his code

src/Farm-Lightbringer.au3
```diff
@@ -80,135 +83,135 @@ Func LightbringerFarmSetup()
...
-	RndSleep(1500)
+	RandomSleep(1500)
 	ActionInteract()
-	RndSleep(1500)
+	RandomSleep(1500)
 
-	If MultipleMoveToAndAggro('First Undead Group', -800, 12000, -1700, 9800) Then Return 1
-	If MultipleMoveToAndAggro('Second Undead Group', -3000, 10900, -4500, 11500) Then Return 1
+	If MultipleMoveToAndAggro('First Undead Group', -800, 12000, -1700, 9800) == $FAIL Or _
+		MultipleMoveToAndAggro('Second Undead Group', -3000, 10900, -4500, 11500) == $FAIL Then Return $FAIL
```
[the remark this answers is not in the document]
Gahais: okay, I will refactor this back. After all this could be dangerous if in some AutoIT version all functions are invoked in if statement, even when if statement was already true (not using lazy evalutation like now). So would still movingaggroing even when previous moveaggro failed

lib/Utils.au3
```diff
@@ -1970,50 +2132,43 @@ Func HasRezMemberAlive()
...
-	Local $count = 0
+	Local $heroes[0]
 	For $heroNumber = 1 To GetHeroCount()
 		Local $heroID = GetHeroID($heroNumber)
 		For $skillSlot = 1 To 8
 			Local $skill = GetSkillbarSkillID($skillSlot, $heroNumber)
 			If IsRezSkill($skill) Then
-				$heroes[$count] = $heroNumber
-				$count += 1
+				_ArrayAdd($heroes, $heroNumber) ;~ addition to dynamic array with automatic resizing
```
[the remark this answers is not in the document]
Gahais: hmm, didn't think about it from the performance point of view. Probably, indeed I noticed some slower performance of fight functions
ok, I will refactor this back. So remove _ArrayAdd in all code places that interact with game in real-time like GetNPCsInRangeOfCoords() too probably?
and keep _ArrayAdd in inventory Management functions?

lib/Utils.au3
Gahais: Hey, below are some notes I made on differences between maps and dictionaries in AutoIT. I also wrote more information in below reply. Correct me if I'm wrong. You can choose which one should be used. Yeah, you had good idea to compress all parameters into map. This way one generic fight function can be used with map with variable parameters

Maps use bracket [] operators to add and access elements. dictionaries use .Add() and .Item() functions to add and access elements. Furthermore, it seems that maps are copied when passed to functions without using ByRef keyword. When ByRef keyword is used for maps then they are indeed passed by reference and original map is modified without copying. In case of dictionaries, these are objects so these will always be passed as reference to functions regardless whether ByRef keyword is used or not. Furthermore for maps ByRef and default arguments probably can't be used simultaneously in function definitions, which seems to be quite an AutoIT limitation. Seems like default Null argument can be used for dictionaries. It would be nice to have default Null argument for options map/dictionary so that function can be called without creating/cloning map/dictionary with some default options map/dictionary used under the hood.
Can convert this options map into a dictionary which should be always passed as reference, so it probably could save some processing time by not copying these maps quite often when calling fight function from moveaggro function

src/Farm-EdenIris.au3
caustic-kronos: Definitely that last loop type if it fits, it's indeed much more elegant.
In your example, $obj2 us a reference to the structure.
So it's safe: we don't edit it, we only read it, and we read it practically as soon as it's been filled so we read valid data.

src/Farm-Lightbringer.au3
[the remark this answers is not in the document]
caustic-kronos: Yeah I think it's safe because AutoIt does lazy evaluation since before it does maps, so even if someone had an old AutoIt version, it would fail because of the maps before it even causes issues for this 🤣 

But a lot of people who write AutoIt bots are not developers and won't understand this syntax, and even for devs like us I think it is clearer to have sequential statements. If they become too many then using a for loop like I saw you do in other places is a good option too

src/Farm-EdenIris.au3
Gahais: ok, you are right, there doesn't seem to be anything dangerous about it. Let's take advantage of this elegant syntax of AutoIT which is quite rare in AutoIT :smile:. I'll refactor all these loops

src/Farm-Lightbringer.au3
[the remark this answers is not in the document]
Gahais: ok, working on it

lib/Utils.au3
caustic-kronos: I have a preference for maps in general, but I suppose it's better to use a dictionary here, just to prevent any issues popping later: there won't be map copy as you said, but also if we want to cascade the options to other functions then we won't have to put ByRef everywhere. So it's probably a good choice.

Since we don't have to use ByRef, that mean that we can refactor :
```
Func WaitAndFightEnemiesInArea($options = Null)
	If IsPlayerAndPartyWiped() Then Return $FAIL

	If $options = Null Then $options = CloneDictMap($Default_MoveAggroAndKill_Options)
```
Into:
```
Func WaitAndFightEnemiesInArea($options = $Default_MoveAggroAndKill_Options)
	If IsPlayerAndPartyWiped() Then Return $FAIL
```
AutoIt has some limitations with default values being variables (have to make sure the variables are properly defined **before** the function is defined. But here it should be fine.

caustic-kronos:
> do You prefer if I do interactive rebase and edit these old commits to reflect your new suggestions

Honestly, anything you're more comfortable with, I don't mind.

Concerning block comments: I see, so it's for aesthetics. I'm fine with it, but I would cut the CS and CE explanations in that case. Even though I agree with you on more descriptive names, in that scenario it's distracting from the comments themselves in my opinion. I think we can either go for :
`; ============================================================================`
Or
`#CS ==========================================================================`
Or no header and keep as it was.

Convention I used before was `;~` on function descriptions, `; ` on comments, and `;` on commented code (no space between code and comment). Honestly it's not that important, but refactoring every comment would be a pain, so keeping previous format might be preferable 😆 

Okay for Attributes_Array and Is(Rarity)Identified👍 

I'm not coding actively these days, not overly motivated, so don't worry about any conflicts and such, you are mostly up to date with your current code. And sure, if you have any other part you want to include (as a different PR though), don't hesitate to. Maybe I could add you as collaborator on the repository if that makes it easier for you.

Visual Studio Code is really good. I used Eclipse and now use IntelliJ Idea at work, both are good but overengineered for AutoIt. Visual Studio Code keeps it simple while still able to provide top of the line utilities. SciTe is just like AutoIt in comparison: it is 20 years behind the times. I don't know CLion but it's JetBrains too so it's probably in the same ballpark as Idea: really good, but too much for AutoIt needs.

Concerning functions not working and such: yes there are a few of those, some might already have been fixed - if you're on the discord, you can take a look at what's happening there, KleuTSchi and Greg are repairing stuff sometimes. I dipped my toes into fixing incorrect memory accesses and solved a function or two myself but I'm still far from understanding it all and doing it easily. I started writing a little guide to get into this, it's in one branch, but I never finished it.

The example of AdlibRegister you found is quite awful 😄 
It's happening way too often (1 second), so it creates many interruptions, and it seems to be only for time, which makes little sense. I tried to keep the AdlibRegister to the minimum, the ones I use are :
- the one that updates the progressBar: that's the one I feel the worst about, it runs every 5s so it's not great already and it's just for the looks
- the one to track group status, it has to run every 10s because more mean we could miss the group death, and less that we could count a single group death several times.
- one for creating map from coordinates, this one I just use as one shot when building bot so no worries
- 3 in OmniFarmer but only one should run at the same time. The one on a too short timer (600ms) I have seen create issues so I don't use it anymore.

As you can see in my other comment, I thought a lot about this Map/Dictionary, and even though I would really want to just use maps, I think you have the right instinct in using Dictionary here. So no worries you can keep it the way it is.

Gahais:
ok, looks like I finished this revision refactor more or less :) I edited master branch with force push so that it has clean commit history. And copied old master branch before revision refactor into this branch old-master-branch-before-revisions-refactor
You can compare these 2 branches to see the refactor if you want, for example using Meld program


these are new commits:
- chore: Update of Credits.md
- chore: Added contribution info to top of files
wrote this contribution information as suggested

- refactor: commented TrackPartyStatus code to prevent main thread interruptions - just in case as you mentioned that can cause undefined problems, this code might not be necessary, when using team status checks functions

2 new commits for convenience which can fit this pull request so that set-up of farms is more robust. Split big functions into modular, expandable functions
- refactor: Added convenient functions for leaving outposts/starting missions to farms
- refactor: Added convenient functions for setting up team to farms

- refactor: moved constants out of WarSupply Farm into GWA2_ID - these skill IDs and map ID probably fit better in GWA2_ID file
- refactor: Changed all loops into elegant For In syntax - to have clean loop syntax, and still safe to use, as suggested
- refactor: Removed _ArrayAdd() calls to improve performance - as suggested because _ArrayAdd() can be slow

- refactor: improvement for GWA code related to attributes
to use this attributes array, For $i = 0 To UBound($Attributes_Array) - 1 in place of For $i = 0 To 44

refactor: Changed many calls to moveaggro into arrays and loops - to avoid using chained If Or _ statements and avoid using many calls to moveaggro functions, but use quite compact and clear arrays of all mobs positions
In Norn and Vanguard farm could also try to put blessings positions into that array and invoke some blessing function `if log == "Taking Blessing"`
And do in single line
	`If MoveAggroAndKillGroups($foes, 1, 100) == $FAIL Then Return $FAIL`
But using array for moving, taking blessing and fighting might be a bit weird imho
Also used first and last index of groups to handle because there seem to quite a limitation in AutoIT that can't pass array literals to functions and anywhere like `function([1,2,3,4,5])`



changed these old commits:
- feat: Added Stygian Gemstone farm - removed unused and untested code related to ranger and leave only code related to mesmer and assassin

- style: Added stylish comment blocks - changed these comment blocks as per suggestion into
`#CS (or CE) ==========================================================================`

- refactor: Dictionary maps for passing options by reference optimization - changed as in your suggestion and seems to work fine :)
```
Func WaitAndFightEnemiesInArea($options = Null)
	If IsPlayerAndPartyWiped() Then Return $FAIL
	If $options = Null Then $options = CloneDictMap($Default_MoveAggroAndKill_Options)
```
Into:
```
Func WaitAndFightEnemiesInArea($options = $Default_MoveAggroAndKill_Options)
	If IsPlayerAndPartyWiped() Then Return $FAIL
```


You can revise that in free time and tell what do you think about it :)

lib/Utils.au3
```diff
@@ -2047,317 +2215,534 @@ Func MoveAvoidingBodyBlock($coordX, $coordY, $timeOut)
...
+	$map[$WarSupply_Seeker_6] 		= 8
+	$map[$WarSupply_Seeker_7] 		= 8
+	$map[$WarSupply_Seeker_8] 		= 8
+	$map[$WarSupply_Scout_1]		= 8
+	$map[$WarSupply_Scout_2]		= 8
+	$map[$WarSupply_Scout_3]		= 8
+	$map[$WarSupply_Scout_4]		= 8
+	$map[$WarSupply_Marksman_1]		= 8
+	$map[$WarSupply_Marksman_2]		= 8
+	$map[$WarSupply_Marksman_3]		= 8
```
caustic-kronos: Is it essential to have a priority for the lower priority mobs ?
For instance, is it that important to kill a knight before a marksman ?
If not, I think we can avoid the less important mobs and reduce a bit amount of lines and complexity here.

lib/Utils.au3
```diff
@@ -2421,7 +2806,9 @@ EndFunc
 ;~ Loads skill template code.
 Func LoadSkillTemplate($buildTemplate, $heroIndex = 0)
 	Local $heroID = GetHeroID($heroIndex)
-	Local $splitBuildTemplate = StringSplit($buildTemplate, '')
+	Local $BuildTemplateChars = StringSplit($buildTemplate, '') ; splitting build template string into array of characters
+	; deleting first element of string array (which has the count of characters in AutoIT) to have string array indexed from 0
+	_ArrayDelete($BuildTemplateChars, 0)
```
caustic-kronos: _ArrayDelete is also another terrible performance function to avoid if possible (we recreate the whole array from scratch when calling it).
Here we're using it in a cosmetic fashion, so I don't think it's worth doing it.

lib/Utils.au3
```diff
@@ -1679,287 +1768,367 @@ EndFunc
...
-	$position[1] = $position[1] / $foes[0]
+	$position[0] = $position[0] / Ubound($foes) ; arithmetic mean calculation for X axis
+	$position[1] = $position[1] / Ubound($foes) ; arithmetic mean calculation for Y axis
 	Return $position
 EndFunc
 
 
 ;~ Get foes in range of the given agent
-Func GetFoesInRangeOfAgent($agent, $range = 0, $condition = null)
+Func GetFoesInRangeOfAgent($agent, $range = Null, $condition = Null)
```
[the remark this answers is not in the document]
Gahais: hmm, good idea :) it should be a quick fix. I will try post a new commit soon
You can also check my message below :)

lib/Utils.au3
```diff
@@ -1984,36 +2153,35 @@ Func FindHeroesWithRez()
...
+	Local Static $Lively_Was_Naomei			= 1222
+	Local Static $Restoration				= 963
+	Local Static $Light_of_Dwayna			= 304
+	Local Static $Rebirth					= 306
+	Local Static $Renew_Life				= 1263
+	Local Static $Restore_Life				= 314
+	Local Static $Resurrect					= 305
+	Local Static $Resurrection_Chant		= 1128
+	Local Static $Unyielding_Aura			= 268
+	Local Static $Vengeance					= 315
```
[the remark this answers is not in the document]
Gahais: hey, also a good idea :) definitely should replace these skill IDs into GWA2_ID
Currently amount of skills is quite few, imho, when GWA2_ID.au3 starts to blow from the amount of skills typed there then can create a new file like GW_SKILL_IDS.au3 and put them all there
it should be a simple fix. I will try post a new commit soon, maybe tomorrow
You can also check my message below :)

src/Farm-FoW.au3
```diff
@@ -42,68 +45,69 @@ Global Const $ID_FoW_Unholy_Texts = 2619
...
+EndFunc
+
+
 ;~ Farm loop
 Func FoWFarmLoop()
-	ResetFailuresCounter()
-	AdlibRegister('TrackGroupStatus', 10000)
+	If GetMapID() <> $ID_Fissure_of_Woe Then Return $FAIL
+	;ResetFailuresCounter()
+	;AdlibRegister('TrackPartyStatus', 10000)
```
caustic-kronos: - [ ] Why disable the TrackPartyStatus ?

If they don't run there is no more management of team death - did you add it somewhere else ?

lib/Utils.au3
Gahais: hmm, I can try writing some custom (but still simplistic) function for deleting any element in array. Maybe it would provide better performance?

lib/Utils.au3
Gahais: ok, good idea. I'll do that :)

lib/Utils.au3
caustic-kronos: There is no clean way to do this in AutoIt - arrays are fixed size, so using _ArrayAdd, _ArrayDelete or ReDim are all reallocating an array. In other words, any custom function that you write to delete an element of an array will have to reallocate (or moving all elements one index before and setting last to null, effectively forcing array's consumer to deal with null values: not great). We don't have overly big arrays, but we do have some functions call very frequently, so if we are not careful with this kind of operation we can make lots of array allocations.
Options to avoid this kind of issue are :
- using an array that is correctly built from scratch: no need for addition, deletion or resizing
- using a list from a library (need to find a good library that implements list properly and not like some kind of terrible implementation over arrays)
- using maps as a list (key = index): this is more costly memory wise, but it allows easy addition and removal from the end of the "list", though removal from anywhere before the end of the list is annoying (have to shift all indices)
- accept that AutoIt, in some cases, have indexes starting at 1 - I think it's okay to do this for some cases, like this one.

src/Farm-FoW.au3
caustic-kronos: Hmm, this is risky.
Problem is that bot can fail in other places, like when opening chests, activating locks, picking loot, or even simply moving. If we want to cover all those cases, we suddenly need to check death in all of those functions.
This will then cause more performance issues because instead of checking once every 10s that the team is doing fine, we are suddenly checking at each move, each moveaggroandkill, each pickUpLoot, etc ... which means we'll retrieve informations from all groups more often and ultimately lose performance.
This change I think should be reverted.

lib/Utils.au3
```diff
@@ -153,7 +154,7 @@ EndFunc
 
 #Region Loot items
 ;~ Loot items around character
-Func PickUpItems($defendFunction = Null, $ShouldPickItem = DefaultShouldPickItem, $range = $RANGE_COMPASS)
+Func PickUpItems($DefendFunction = Null, $ShouldPickItem = DefaultShouldPickItem, $range = $RANGE_COMPASS)
```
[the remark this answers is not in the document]
Gahais: ok, of course, I can refactor this quickly :)

src/Farm-Lightbringer.au3
```diff
@@ -130,15 +130,50 @@ Func FarmTheSulfurousWastes()
...
+		[-19500, -8500, 'Seventh Margonite Group 2'], _
+		[-22000, -9400, 'Temple Monolith Group 1'], _
+		[-23000, -10600, 'Temple Monolith Group 2'], _
+		[-22800, -13500, 'Temple Monolith Group 3'], _
+		[-19500, -13100, 'Temple Monolith Group 4'], _
+		[-18000, -13100, 'Temple Monolith Group 5'], _
+		[-18000, -13100, 'Margonite Boss Group'] _
+	]
+
+	If MoveToAndAggroGroups($foes, 1, 4) == $FAIL Then Return $FAIL
```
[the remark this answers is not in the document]
Gahais: hmm, you mean setting up small arrays before call of MoveToAndAggroGroups() function? It seems to be doable.
I think the biggest problem in situation like this is limitation of AutoIT that can't pass array literals to functions like 2D arrays with unlimited number of elements, like any number of coordinates into such function.
This would be possible and probably easier to do in other languages :)

lib/Utils.au3
caustic-kronos: I agree, on this function it does not matter that much, but what I meant is that this solution is not clean and so we can't keep convention of having all arrays starting at 0 in some cases.
But it's fine here, you're right.

src/Farm-FoW.au3
caustic-kronos: Sure, if we can dodge having to run TrackPartyStatus on a AdlibRegister, so much the better.
But let's keep this change out of this PR - I suspect things will break to some degree so I'd prefer having it in another just to be safe, and everything else in this PR looks ok already.

src/Farm-Lightbringer.au3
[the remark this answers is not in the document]
caustic-kronos: Yes, AutoIt only accepts arrays defined on a variable declaration. It's pretty weird honestly.
And yes I meant several small arrays, but I'm not sure it's much much better. It's good as is I believe.

src/Farm-FoW.au3
Gahais: ok, you are the creator and maintainer of this repository, you can decide :)
ok, so I will drop this commit "refactor: Improvements for checking party status", so that the code is reverted back to using TrackPartyStatus on a AdlibRegister.
I can readd this commit in next pull request, but if you don't want it then I can just keep it in my local code, and not commit that to your repository

caustic-kronos:
Trying out the Stygian farmer:
- I had one crash on the first (successful) run (the exact same issue you described), so I restarted with debug mode enabled
- but since then I have had only failures on the bot: the mobs are de-aggroing in the middle of the fight, so they receive uneven damage, some die, some don't, and the next spawn behind the character is triggered while the bot is still busy trying to clear the leftovers of mobs.

I see the wiki farm description mentions:
"Taking a 1-2 second break between cycles should avoid the far group breaking off. Use Wastrel's Worry on them if they do begin to run."
So this might be the way to solve that issue.

Gahais:
Thanks for merging the code :)

ok, I tried improving stygian gemstone farm a bit, checked the wiki article
There was probably one more problem that these mobs use grenth's balance skill that boosts health to some mobs randomly with uneven distribution of health
So I had an idea to remove wastrel's worry skill and use only wastrel's demise skill which makes small damage which should not scare away the mobs :)
And to use shadow form with deadly paradox instead of shadow sanctuary and wastrel's worry. So there is a permanent shadow form which blocks grenth's balance skill. New builds are automatically loaded
Should wear weapon +20% longer enchantments
And added 10 seconds wait-out before casting this wastrel skill so that all mobs have time to come and gather in front of player

I committed that improvement to this commit 78eaa23. You can cherry pick it somehow or apply below expandable git patch

if stygian gemstone farm is too bad to use then we can ignore it and remove it in the next pull request
I may do second pull request soon


<details>
<summary>Click to expand git patch code</summary>

```
Subject: [PATCH] fix: prevent mobs from deaggroing in stygian farm
---
Index: src/Farm-GemstoneStygian.au3
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/Farm-GemstoneStygian.au3 b/src/Farm-GemstoneStygian.au3
--- a/src/Farm-GemstoneStygian.au3	(revision f6f80d8a3160bd588ecce8381229b69d10b7fd5f)
+++ b/src/Farm-GemstoneStygian.au3	(revision 78eaa237bd3d93ec311f507037c81335712a4033)
@@ -29,13 +29,13 @@
 
 #Region Configuration
 ; === Build ===
-Global Const $StygianAssasinSkillBar = "OwVTI4h9X6mSGYFct0E4uM0ZCCA"
-Global Const $StygianMesmerSkillBar = "OQdUASBPmfS3UyArgrlmA3lhOTQA"
+Global Const $StygianAssasinSkillBar = "OwVT8ZBPGiHRn5mat0E4uM0ZCC"
+Global Const $StygianMesmerSkillBar = "OQdTI4x8ZiHRn5mat0E4uM0ZCC"
 Global $StygianPlayerProfession = $ID_Mesmer ; global variable to remember player's profession in setup and avoid creating Dll structs over and over
 
-Global Const $Stygian_WastrelsDemise		= 1
-Global Const $Stygian_WastrelsWorry			= 2
-Global Const $Stygian_ShadowSanctuary		= 3
+Global Const $Stygian_DeadlyParadox			= 1
+Global Const $Stygian_ShadowForm			= 2
+Global Const $Stygian_WastrelsDemise			= 3
 Global Const $Stygian_Mindbender			= 4
 Global Const $Stygian_Channeling			= 5
 Global Const $Stygian_DwarvenStability		= 6
@@ -44,7 +44,9 @@
 #EndRegion Configuration
 
 ; ==== Constants ====
-Global Const $GemstoneStygianFarmInformations = '' & @CRLF _
+Global Const $GemstoneStygianFarmInformations = 'For best results, have :' & @CRLF _
+	& ' - Armor with HP runes and 5 blessed insignias (+50 armor when enchanted)' & @CRLF _
+	& ' - Weapon of Enchanting (20% longer enchantments duration) to make Shadow Form permanent' & @CRLF _
 	& 'You can run this farm as Assassin or Mesmer. Bot will set up build automatically for these professions' & @CRLF _
 	& 'This bot farms stygian gemstones (1 of 4 types) in Stygian Veil location' & @CRLF _
 	& 'Player needs to have access to Gate of Anguish outpost which has exit to Stygian Veil location' & @CRLF _
@@ -56,8 +58,8 @@
 	& 'For Mesmer and Assassin this bot works by exploitation of Guild Wars bug in pathing AI, which causes mobs to not attack player' & @CRLF _
 	& 'Caution, it looks like because of this bug exploitation the guild wars client crashes very soon' & @CRLF
 ; Average duration ~ 5 minutes
-Global Const $GEMSTONE_STYGIAN_FARM_DURATION = 5 * 60 * 1000
-Global Const $MAX_GEMSTONE_STYGIAN_FARM_DURATION = 10 * 60 * 1000
+Global Const $GEMSTONE_STYGIAN_FARM_DURATION = 8 * 60 * 1000
+Global Const $MAX_GEMSTONE_STYGIAN_FARM_DURATION = 16 * 60 * 1000
 Global $GemstoneStygianFarmTimer = Null
 Global $GEMSTONE_STYGIAN_FARM_SETUP = False
 Global Const $Stygians_Range_Short = 800
@@ -125,7 +127,7 @@
 		Warn("You need to run this farm bot as Assassin or Mesmer")
 		Return $FAIL
     EndIf
-	ChangeWeaponSet(1) ; change to other weapon slot or comment this line if necessary
+	;ChangeWeaponSet(1) ; change to other weapon slot or comment this line if necessary
 	Sleep(500)
 EndFunc
 
@@ -240,27 +242,22 @@
 Func KillStygianMobsUsingWastrelSkills()
 	If IsPlayerDead() Then Return $FAIL
 	Local $me, $target, $distance
-	RandomSleep(500)
+	Sleep(10000) ; waiting for all mobs to come
 
 	While CountFoesInRangeOfAgent(GetMyAgent(), $Stygians_Range_Long) > 0 And IsPlayerAlive()
 		If TimerDiff($GemstoneStygianFarmTimer) > $MAX_GEMSTONE_STYGIAN_FARM_DURATION Then Return $FAIL
+		StygianCheckSFBuffs()
 		$me = GetMyAgent()
 		$target = GetNearestEnemyToAgent(GetMyAgent())
 		ChangeTarget($target)
-		If GetHealth(GetMyAgent()) < 300 And IsRecharged($Stygian_ShadowSanctuary) And GetEnergy() > 5 Then UseSkillEx($Stygian_ShadowSanctuary)
-		If IsRecharged($Stygian_Channeling) And GetEnergy() > 5 Then
+		If IsRecharged($Stygian_Channeling) And Not IsRecharged($Stygian_ShadowForm) And GetEnergy() > 5 Then
 			UseSkillEx($Stygian_Channeling)
 			RandomSleep(100)
 		EndIf
 		$distance = GetDistance($me, $target)
-		If IsRecharged($Stygian_WastrelsDemise) And GetEnergy() > 5 And $distance < $Stygians_Range_Short Then
+		If Not GetHasHex($target) And IsRecharged($Stygian_WastrelsDemise) And Not IsRecharged($Stygian_ShadowForm) And GetEnergy() > 5 And $distance < $Stygians_Range_Short Then
 			UseSkill($Stygian_WastrelsDemise, $target)
-			RandomSleep(300)
-		EndIf
-		$distance = GetDistance($me, $target)
-		If IsRecharged($Stygian_WastrelsWorry) And GetEnergy() > 5 And $distance < $Stygians_Range_Short Then
-			UseSkill($Stygian_WastrelsWorry, $target)
-			RandomSleep(300)
+			Sleep(1000)
 		EndIf
 		RandomSleep(100)
 	WEnd
@@ -270,7 +267,17 @@
 EndFunc
 
 
-Func StygianCheckBuffs()
+Func StygianCheckSFBuffs()
+	If IsPlayerDead() Then Return $FAIL
+	If IsRecharged($Stygian_DeadlyParadox) And IsRecharged($Stygian_ShadowForm) And GetEnergy() >= 20 Then
+		UseSkillEx($Stygian_DeadlyParadox)
+		UseSkillEx($Stygian_ShadowForm)
+	EndIf
+	Return IsPlayerAlive()? $SUCCESS : $FAIL
+EndFunc
+
+
+Func StygianCheckRunBuffs()
 	If IsPlayerDead() Then Return $FAIL
 	If IsRecharged($Stygian_DwarvenStability) And GetEnergy() > 5 Then UseSkillEx($Stygian_DwarvenStability)
 	If IsRecharged($Stygian_Dash) And GetEnergy() > 5 Then UseSkillEx($Stygian_Dash)
@@ -297,22 +304,22 @@
 	Local $ChatStuckTimer = TimerInit()
 	Move($destinationX, $destinationY, $random)
 	While IsPlayerAlive() And GetDistanceToPoint(GetMyAgent(), $destinationX, $destinationY) > $random * 1.5
-	    StygianCheckBuffs()
+	    StygianCheckRunBuffs()
 		;TargetNearestEnemy()
 	    $target = GetNearestEnemyToAgent(GetMyAgent())
 	    ChangeTarget($target)
 		RandomSleep(50)
 		$me = GetMyAgent()
 		If GetIsDead($me) Then Return $FAIL
-		StygianCheckBuffs()
+		StygianCheckRunBuffs()
 		If Not IsPlayerMoving() Then
-			StygianCheckBuffs()
+			StygianCheckRunBuffs()
 			$blocked += 1
 			If $blocked < 5 Then
 				Move($destinationX, $destinationY, $random)
 			ElseIf $blocked < 30 Then
 				$angle += 40
-				StygianCheckBuffs()
+				StygianCheckRunBuffs()
 				Move(DllStructGetData($me, 'X')+300*sin($angle), DllStructGetData($me, 'Y')+300*cos($angle))
 			EndIf
 		Else
@@ -323,7 +330,7 @@
 				EndIf
 				$blocked = 0
 			EndIf
-			StygianCheckBuffs()
+			StygianCheckRunBuffs()
 			If GetDistance($me, $target) > 1100 Then ; target is far, we probably got stuck.
 				If TimerDiff($ChatStuckTimer) > 3000 Then ; dont spam
 					SendChat("stuck", "/")

**Comment by Gahais** [comment]
I'm sorry for breaking up Froggy and SoO farms. I overlooked this one.
I see that you have already fixed these farms though, in recent 3 commits.
ok, we can keep Gemstone Stygian farm. I continue with second pull request

**Comment by Gahais** [comment]
hi, sorry for breaking up Pongmei farm skill targetting
I think I know what can be the problem
I tried to make GetTargetForDeathsCharge() function similar to GetNPCInTheBack(), that unselected agent/npc is Null instead of 0, like using Null for structs, 0 for integers
in line 307 in Pongmei farm I replaced 
	`Return 0`
with
	`Return Null`
So now probably should also replace in this line 209 in Pongmei farm
			`If $target <> 0 Then UseSkillEx($Pongmei_DeathsCharge, $target)`
with
			`If $target <> Null Then UseSkillEx($Pongmei_DeathsCharge, $target)`

In AutoIT, it seems that Null is different from 0

below is expandable git patch that changes this line


<details>
<summary>Click to expand git patch</summary>

```
Subject: [PATCH] fix: fix for Pongmei farm
---
Index: src/Farm-Pongmei.au3
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/Farm-Pongmei.au3 b/src/Farm-Pongmei.au3
--- a/src/Farm-Pongmei.au3	(revision d205f3edf57daaba86b6d52da77008f6d1ddb066)
+++ b/src/Farm-Pongmei.au3	(revision 7c32a93f54602fab2c1d93dcee8b7a717252692e)
@@ -206,7 +206,7 @@
 
 		If GetEnergy() >= 5 And IsRecharged($Pongmei_DeathsCharge) Then
 			Local $target = GetTargetForDeathsCharge($X, $Y, 700)
-			If $target <> 0 Then UseSkillEx($Pongmei_DeathsCharge, $target)
+			If $target <> Null Then UseSkillEx($Pongmei_DeathsCharge, $target)
 		EndIf
 
 		If GetEnergy() >= 20 And IsRecharged($Pongmei_ShadowForm) And AreFoesInFront($X, $Y) Then
```

**Comment by Gahais** [comment]
I think that Vanguard bot gets stuck on this line 2474 in Utils.au3, loop that uses skills until player or target is deceased
`While Not IsPlayerDead() And Not GetIsDead(GetCurrentTarget()) And $target <> Null And DllStructGetData($target, 'ID') <> 0`
My theory is that in this specific case same agent with same ID has initially allegiance enemy == 3 which later changes to allegiance npc == 6, similarly to these fights with Norns near Norn shrines
Maybe we could change that line to this one, to keep looping while target's allegiance is enemy == 3
`While Not IsPlayerDead() And Not GetIsDead(GetCurrentTarget()) And $target <> Null And DllStructGetData($target, 'ID') <> 0 And DllStructGetData($target, 'Allegiance') == 3`

This may work, I haven't tested it out though

below is expandable git patch that changes this line
I committed this patch to fix branch just in case it may work


<details>
<summary>Click to expand git patch for Vanguard farm</summary>

```
Subject: [PATCH] fix: fix for Vanguard farm stuck
---
Index: lib/Utils.au3
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/lib/Utils.au3 b/lib/Utils.au3
--- a/lib/Utils.au3	(revision 950bd4061c602d63207c98a8946cf12dcdc62c97)
+++ b/lib/Utils.au3	(revision 30d76f2f3f8ad9ff5c00914336f541f69e4c3609)
@@ -2471,7 +2471,7 @@
 
 			Local $i = 0 ; index for iterating skills in skill bar in range <1..8>
 			; casting skills from 1 to 8 in inner loop and leaving it only after target or player is dead
-			While Not IsPlayerDead() And Not GetIsDead(GetCurrentTarget()) And $target <> Null And DllStructGetData($target, 'ID') <> 0
+			While Not IsPlayerDead() And Not GetIsDead(GetCurrentTarget()) And $target <> Null And DllStructGetData($target, 'ID') <> 0 And DllStructGetData($target, 'Allegiance') == 3
 				$i = Mod($i, 8) + 1 ; incrementation of skill index and capping it by number of skills, range <1..8>
 				If $skillsMask <> Null And $skillsMask[$i-1] == False Then ContinueLoop ; optional skillsMask indexed from 0, tells which skills to use or skip
 

```

PR #46, opened 2025-12-19
fix:		GW Reforged fix - not complete yet

caustic-kronos:
Will need to be updated with additional fixes for yesterday's update

### DIFF
--- a/BotsHub.au3
+++ b/BotsHub.au3
@@ -73,6 +73,7 @@
 #include 'src/Farm-SpiritSlaves.au3'
 #include 'src/Farm-SunspearArmor.au3'
 #include 'src/Farm-Tasca.au3'
+#include 'src/Farm-TestSuite.au3'
 #include 'src/Farm-Vaettirs.au3'
 #include 'src/Farm-Vanguard.au3'
 #include 'src/Farm-Voltaic.au3'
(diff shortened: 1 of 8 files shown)


caustic-kronos:
All fixed, except selling materials at material merchant it seems.
Merging.

jeepgeek93-tech:
also Follower has some issues https://ibb.co/9mZZNmVM per what another player had said

jeepgeek93-tech:
@Gahais , for the selling issue it seems that others are also commenting in the elitepvpers forum having the same issue. It causes a generic client crash when trying to buy ecto and sometimes the bot does not port to EotN and thinks it in the outpost. 
As for the follower bot, I am unsure about the second issue the player was having. I tried running it myself in pre searing with just 1 other account and it was giving me the same error message.

Gahais:
hey @jeepgeek93-tech
ok, I have a fix for both issues :)

Issue that bot does not port to EotN and thinks it's in the outpost
This problem is caused by different festival map type of EotN
It looks like each character needs to visit city decorated during events first before being able to travel automatically to that city decorated during events using bots
Otherwise that city is considered an unknown outpost to which bot can't travel even when that city was visited before festival event
So just visit outpost decorated with festival with every character you possess in your accounts before using any bots


Issue in the Follower bot crashing on subscript index exceeding 8 party size
It looks like this error will happen if player is in outposts, where there is more than 8 players, like in Kamadan
Because other players in town are also matched as team members even when they are not in team, other players in town have the same Allegiance = 1 and the same TypeMap containing 0x20000 bit like team members and there doesn't seem to be an easy way to make difference between other players and team members, everybody seems to have team ID = 0 regardless if they are in team or not, that's weird

So I recommend this fix
Add below 4 lines before line 
`$fullParty[$partySize] = $agent`
just after this line
`Func GetParty($agents = Null)`
	
```
	If GetInstanceType() <> $ID_Explorable Then
		Local $outpostTeam[1] = [GetMyAgent()]
		Return $outpostTeam
	EndIf
```
	
I applied this fix into my pull request which I will commit soon
When calling GetParty() function, check if current map is an outpost and if yes then return team consisting only of player. Follower bot probably makes sense only in explorable maps after all

You can tell about these fixes to these other players, spread the message :)

jeepgeek93-tech:
ok well I have not been able to test anything now because everything now crashes the game or comes back saying "you are unable to enter that outpost" :(

PR #59, opened 2026-01-08
LDOA: fixed: retaking Charr at the Gates quest for lvl 2-10

n1kn4x:
The current bot does not retake the Charr at the Gate quest for lvl 2-10.
This has been fixed by adding a simple SetupCharrAtTheGateQuest() call to the beginning of the lvl 2-10 routine.


caustic-kronos (approved):
Thanks for the fix, merging :D

PR #91, opened 2026-01-26
Adding additional Kurzick VQ

NorthboundCoder:
Hi,

No need to accept if you're not interested but decided to map another Kurzick VQ Bot. 

Bot only takes around 25 minutes to run and fully VQs Drazach. Figured it would be a good alternative to the Ferndale


Ian-Coghlan:
@caustic-kronos I think I've covered everything in my latest commits. When you can, please review and let me know :)

PR #92, opened 2026-01-26
Secret Lair of the Snowmen

NorthboundCoder:
Decided to crack my hand at a full new bot.

10 Successful Runs in NM, 6 in HM. 

Requires you to have already done the map pieces but will then pick up the quest, enter and run as needed.

### DIFF
--- a/BotsHub.au3
+++ b/BotsHub.au3
@@ -78,6 +78,7 @@
 #include 'src/utilities/TestSuite.au3'
 #include 'src/vanquishes/Asuran.au3'
 #include 'src/vanquishes/Kurzick.au3'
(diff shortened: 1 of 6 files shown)


caustic-kronos (dismissed):
Pretty neat !
I like it 😄 
There's not much to improve/correct here.
I left two small comments but overall it's all good 👍

Ian-Coghlan:
Missed a flag in previous.

Setting the farm setup back to false so that it will loop through checking for the quest, quest reward and looping in and out.

caustic-kronos (dismissed):
Thanks for correcting, let's merge 😄

PR #114, opened 2026-02-10
fixed war_preparations quest id and lady althea not in compass range

n1kn4x:
fixed some issues with the ldoa bot in InitialSetupLDOA.

Currently the bot is not functional though, since the TakeQuestReward function seems to not work properly.


src/titles/LDOA.au3
```diff
@@ -98,23 +98,24 @@ Func InitialSetupLDOA()
 	MoveTo(10399, 318)
 	MoveTo(11004, 1409)
 	Local $questNPC = GetNearestNPCToCoords(11683, 3447)
-	TakeQuest($questNPC, $ID_QUEST_WAR_PREPARATIONS, $ID_DIALOG_ACCEPT_QUEST_WAR_PREPARATIONS)
+	TakeQuest($questNPC, $ID_QUEST_WAR_PREPARATIONS3, $ID_DIALOG_ACCEPT_QUEST_WAR_PREPARATIONS)
```
caustic-kronos: This is actually more complicated than it looks.
There is a different quest ID per class, and maybe different dialogs as well.
Update code like so:

```
	Local $questID
	Switch DllStructGetData(GetMyAgent(), 'Primary')
		Case $ID_MESMER
			$questID = $ID_QUEST_WAR_PREPARATIONS_MESMER
		Case $ID_NECROMANCER
			$questID = $ID_QUEST_WAR_PREPARATIONS_NECROMANCER
		Case $ID_ELEMENTALIST
			$questID = $ID_QUEST_WAR_PREPARATIONS_ELEMENTALIST
		Case $ID_MONK
			$questID = $ID_QUEST_WAR_PREPARATIONS_MONK
		Case $ID_WARRIOR
			$questID = $ID_QUEST_WAR_PREPARATIONS_WARRIOR
		Case $ID_RANGER
			$questID = $ID_QUEST_WAR_PREPARATIONS_RANGER
	EndSwitch
	TakeQuest($questNPC, $questID, $ID_DIALOG_ACCEPT_QUEST_WAR_PREPARATIONS)
```
And in GWA2_ID_Quests we need to update the names (1,2,3,4,5,6) by:
```
Global Const $ID_QUEST_WAR_PREPARATIONS_MESMER = 217
Global Const $ID_QUEST_WAR_PREPARATIONS_NECROMANCER = 218
Global Const $ID_QUEST_WAR_PREPARATIONS_ELEMENTALIST = 219
Global Const $ID_QUEST_WAR_PREPARATIONS_MONK = 220
Global Const $ID_QUEST_WAR_PREPARATIONS_WARRIOR = 221
Global Const $ID_QUEST_WAR_PREPARATIONS_RANGER = 222
```

We might have to check dialogs too.

caustic-kronos (approved):

PR #124, opened 2026-02-16
Feat/drazach refactor

NorthboundCoder:
Hey,

I've improved the Drazach Farm. I've added a loop for each Shrine, as well as a reset so if we rez at a shrine, we'll follow back through the same path as previously completed. 

I haven't experienced any wall running or anything since introducing. 

Thanks,
North

### DIFF
--- a/src/vanquishes/Kurzick2.au3
+++ b/src/vanquishes/Kurzick2.au3
@@ -23,9 +23,6 @@
 #include '../../lib/GWA2_ID.au3'
 #include '../../lib/Utils.au3'
 
-; Possible improvements :
-; - noticed some scenarios where map is not cleared - check whether this can be fixed by adding a few additional locations
-
 Opt('MustDeclareVars', True)
 
 ; ==== Constants ====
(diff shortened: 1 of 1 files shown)


src/vanquishes/Kurzick2.au3
```diff
@@ -94,97 +141,291 @@ Func VanquishDrazach()
...
+
+	Return (Not IsRunFailed())
+EndFunc
+
+
+; =====================================================================
+; Phase 1 -> Shrine 2
+; NOTE: Every StepTo(...) MUST be checked for $FAIL and Return immediately.
+; =====================================================================
+Func FarmToSecondShrine()
```
caustic-kronos: Moving away from an array of position is a clear downgrade, in my opinion.
Having separate functions per shrine is also a mistake I believe (repetition of the same logic).

In SoO and Bogroot it's done that way because every level is very clearly different from the others (torches, interactions, chests). And even then, it would probably be a good move to tidy that into one or several arrays of positions.

But here, I think it's not a correct design choice.

caustic-kronos (changes requested):
Hey Northbound, thanks for the PR !

On this one I’m going to be more critical because several changes affect the overall design, not just implementation details.

My main concern is that we’re moving away from a generic, data-driven structure toward more specific, hardcoded control flow. The previous approach relied on position arrays and shared movement logic, which kept responsibilities clear and minimized duplication. This version introduces shrine-specific functions, additional wrappers, and embedded state checks that expand function scope and fragment the control flow.

The result is:
- less generic behavior
- more repetition of similar logic
- blurred ownership of run state vs. movement logic
- reduced readability of the main execution flow

Overall, this feels like a step back in terms of abstraction and maintainability. I detailed the specific points in the comments below.

NorthboundCoder:
Okay cool, happy for you to be as critical :)

I'll review your feedback and see what I can do!

caustic-kronos:
@NorthboundCoder I will close the PRs - feel free to reopen them or make new ones if you get back on them.
🕺

PR #129, opened 2026-02-20
(fix): Minor tweaks to UW & Vaettirs 

buddyleex:
Underworld: Added longer sleep for mindblades at spot 4, set flag fight function for spot 8 near spawning pools.
GWA2_ID: Added Phantom Chest gadget ID to chest array.
Vaettirs: Extend waypoint beyond portal to ensure character passes through portal. Current coordinates can leave character unzoned, and then stay in Bjora Marches.


src/farms/Vaettirs.au3
```diff
@@ -615,11 +615,13 @@ Func RezoneToJagaMoraine()
 		EndIf
 	WEnd
 	MoveTo(15600, -20500)
-	Move(15865, -20531)
+	Move(15865, -20531, 30)
```
caustic-kronos:
```suggestion
```

Can be safely removed - since Move call behaves like async code, having two Move one after the other is equivalent to only having the second.

PR #136, opened 2026-02-28
feat: Underworld Expansion pt2

buddyleex:
- Added The Four Horsemen quest (w/ caveats see below)
- Added The Escort of Souls quest
- Added Unwanted Guests quest
- Improved Servants of Grenth, Imprisoned Spirits, and Demon Assassin quests
- Improved area clearing of Frozen Wastes, Chaos Planes, Bone Pits, and Twin Serpent Mountain
- Added team setup for combo of Rt/A or A/Rt which is required for The Four Horsemen quest
- Added relevant model IDs for Underworld farm
- Added Underworld priority mob arrays

### DIFF
--- a/src/missions/Underworld.au3
+++ b/src/missions/Underworld.au3
@@ -1179,9 +1179,6 @@ Func ClearTwinSerpentMountains()
 	$optionsTwinSerpentMountains.Item('fightRange') = $RANGE_EARSHOT * 1.25
 	MoveAggroAndKillSafeTraps(-8150, -4800, '', $optionsTwinSerpentMountains)
 	CommandAll(-7988, -4615)
-	MoveTo(8220, 5202)
-	Sleep(5000)
-	CancelAll()
 	MoveTo(-8317, -5353)
 	Sleep(5000)
 	CancelAll()
@@ -1189,6 +1186,7 @@ Func ClearTwinSerpentMountains()
 	Sleep(5000)
 	CommandAll(-8317, -5353)
 	PickUpItems(Null, DefaultShouldPickItem, $RANGE_EARSHOT)
+	CancelAll()
 	
 	Return IsPlayerOrPartyAlive() ? $SUCCESS : $FAIL
 EndFunc
@@ -1398,7 +1396,7 @@ Func UnwantedGuests($reaper)
 		MoveAggroAndKillSafeTraps(5371, 300)
 		MoveAggroAndKillSafeTraps(2753, 2853)
 		MoveTo(1850, 2410)
-		TrackVengefulAatxes(-2424, 1767, 'away', $RANGE_COMPASS, $RANGE_EARSHOT)
+		TrackVengefulAatxes(-1882, 1820, 'away', $RANGE_COMPASS, $RANGE_EARSHOT)
 		RandomSleep(5000)
 		Info('Aggro killable mobs away from Vengeful Aatxe path')
 		CommandAll(1850, 2410)
@@ -1413,7 +1411,7 @@ Func UnwantedGuests($reaper)
 		KillFoesInArea($optionsUnwantedGuests)
 		CommandAll(1850, 2410)
 		RandomSleep(5000)
-		TrackVengefulAatxes(-2424, 1767, 'away', $RANGE_COMPASS, $RANGE_EARSHOT)
+		TrackVengefulAatxes(-1882, 1820, 'away', $RANGE_COMPASS, $RANGE_EARSHOT)
 		CancelAll()
 		MoveAvoidingBodyBlock(-214, 2775)
 		MoveAvoidingBodyBlock(267, 3575)
@@ -1514,7 +1512,6 @@ Func TheFourHorsemen($reaper)
 	Sleep(5000)
 	TakeQuest($reaper_ChaosPlanes, $ID_QUEST_THE_FOUR_HORSEMEN, 0x806A01, 0x806A03)
 	UseSkillEx($UNDERWORLD_RECALL, $reaper_ChaosPlanes)
-	ToggleQuestWindow()
 	Info('Time to Kite the Four Horsemen.')
 	
 	Local $steps[17][4] = [ _
@@ -1539,7 +1536,6 @@ Func TheFourHorsemen($reaper)
 
 	Local $index = 0
 	Local $stepCount = UBound($steps)
-	AbortFourHorsemenKite(True)
 
 	While IsPlayerAlive() And $index < $stepCount And Not AbortFourHorsemenKite()
 		MoveTo($steps[$index][0], $steps[$index][1])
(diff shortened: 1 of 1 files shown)


lib/Utils-Agents.au3
```diff
@@ -836,5 +836,10 @@ Func CreateMobsPriorityMap()
 	Local $ldoaMobs =				[	$ID_BANDIT_RAIDER,	$ID_BANDIT_FIRESTARTER]
 	Local $ldoaMobsPriorities =		[	0,					1]
 	AddToMapFromArrays($priorityMap, $ldoaMobs, $ldoaMobsPriorities)
+	Local $underWorldMobs =				[	$ID_KEEPER_OF_SOULS,	$ID_VENGEFUL_AATXE,		$ID_TERRORWEB_DRYDER,			$ID_TERRORWEB_DRYDER_SILVER,		$ID_SMITE_CRAWLER, _
+											$ID_BANISHED_DREAM_RIDER,	$ID_MINDBLADE_SPECTRE,	$ID_BLADED_AATXE,	$ID_SKELETON_OF_DHUUM]
+	Local $underWorldMobsPriorities =		[	0,							999,							3,							4,								3, _
+											1,							5,								3,							2]
```
caustic-kronos: Would probably be better to order them by priority so we can instantly see which ones have the highest priority.

src/missions/Underworld.au3
```diff
@@ -308,7 +397,7 @@ Func WrathfulSpirits()
 	; Take Quest Wrathful spirits from the Reaper and complete
 	Local $reaper = GetNearestNPCToCoords(-13211, 5322)
 	If Not $ATTEMPT_REAPER_QUESTS Or Not $ENABLE_WRATHFUL_SPIRITS Then
-		Info('Skipping Wrathful Spirits Quest as per settings')
+		Info('Skipping ' & $QUEST_NAMES_FROM_IDS[$ID_QUEST_WRATHFUL_SPIRITS] &' quest per settings.')
```
caustic-kronos: I don't think it's very useful to replace the name of the quest with a dynamic way to get it here, since it's always the same here anyway.

src/missions/Underworld.au3
```diff
@@ -478,39 +571,42 @@ Func ServantsOfGrenth()
...
 	Info('Killing waves of Dryders and Skeletons')
-	While Not IsQuestReward($ID_QUEST_SERVANTS_OF_GRENTH)
-		MoveAggroAndKill(2807, 19907, '', $optionsFrozenWastes)
-		RandomSleep(250)
-		MoveAggroAndKill(2200, 19668, '', $optionsFrozenWastes)
-		RandomSleep(250)
-		If Not IsPlayerOrPartyAlive() Then
-			Info('Quest Failed: ' & $QUEST_NAMES_FROM_IDS[$ID_QUEST_SERVANTS_OF_GRENTH])
-			Return $FAIL
+	While IsPlayerOrPartyAlive()
```
caustic-kronos: The loop should be driven by the quest state, since quest completion is the actual goal of this block.

Not IsQuestReward(...) clearly expresses the intended invariant: keep executing until the quest is complete. Player death is a failure condition and is therefore correctly handled as an early return.

Switching the loop condition to IsPlayerOrPartyAlive() makes survival the primary driver of control flow, which is not the objective of the logic. It also splits success and failure handling across different structural levels and introduces an unnecessary ExitLoop.

For clarity, correctness, and maintainability, the loop predicate should represent the goal state, not a failure guard.

This is valid for all other loops - please revert.

src/missions/Underworld.au3
```diff
@@ -1290,3 +1333,367 @@ Func DemonAssassin()
...
+	CommandHero(1, 11076, -17974)
+	CommandHero(2, 11563, -17492)
+	CommandHero(3, 11744, -17898)
+	CommandHero(4, 10867, -17746)
+	CommandHero(5, 11420, -17833)
+	CommandHero(6, 11240, -17213)
+	CommandHero(7, 10894, -17430)
+
+	Info('Final Stand against the Four Horsemen at Reaper.')
+	While IsPlayerOrPartyAlive()
```
caustic-kronos: Same problem as all other while loops - to make it clear what this while loop does, you should have as its condition its "goal", the thing that makes it exit.
Here it looks like the goal is to die, and the exception is to finish the quest - which is why it would be preferable to reverse the operations and make the goal to finish the quest and the exception to die.

caustic-kronos (changes requested):
Hey @buddyleex you really want to kill Dhuum 😆 
I added some comments, there are a few issues here and there.
The most glaring one I think is that while loops are not used properly and there are some misunderstanding in how they work (i.e: the condition is only checked once per loop).
https://www.autoitscript.com/autoit3/docs/keywords/While.htm

src/missions/Underworld.au3
```diff
@@ -141,6 +189,47 @@ Func TeleportBackToChaosPlains($reaper)
...
+	Sleep(1000)
+	GoToNPC($reaper)
+	Sleep(1000)
+	Dialog(0x7F)
+	Sleep(1000)
+	Dialog(0x87)
+	Sleep(1000)
+	Dialog(0x8E)
+	Sleep(1000)
+EndFunc
```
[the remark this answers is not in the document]
buddyleex: That's a solid suggest, I refactored under 9fd3ff9fd08974228c2edec83fedcb0970528a3c hopefully it's all right hehe especially how I think that initialID will work by sending it Null to ignore the "initial" dialog entry.

src/missions/Underworld.au3
buddyleex: I think I got all references like this refactored now in 631501f5d9c87aac0e99c0c2662e5bfe9e9ae856

src/missions/Underworld.au3
```diff
@@ -1290,3 +1333,367 @@ Func DemonAssassin()
...
+	UseSkillEx($UNDERWORLD_ARMOR_OF_UNFEELING)
+	DropBuff($ID_RECALL, GetMyAgent())
+	Sleep(5000)
+	TakeQuest($reaper_ChaosPlanes, $ID_QUEST_THE_FOUR_HORSEMEN, 0x806A01, 0x806A03)
+	UseSkillEx($UNDERWORLD_RECALL, $reaper_ChaosPlanes)
+	MoveTo(7468, -19050)
+	Sleep(7000)
+
+	Local $myHealth = DllStructGetData(GetMyAgent(), 'HealthPercent')
+	While $myHealth > 0.3
```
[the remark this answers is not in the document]
buddyleex: I really want to end the loop when health falls, so I've revamped the loop if you can please check out and see if there might be a better way? a9c64479217d68725f98f98c2b53ae593fe3b7b6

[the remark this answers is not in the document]
caustic-kronos: Probably need to make a global util to get objective yes.
Is your AbortFourHorsemenKite working ? It looks a bit sus - AI is good for refactoring but for making proper memory functions it's sketchy.

caustic-kronos (commented):
Hey @buddyleex I added some additional fixes and committed this morning the function you need to read the objectives.
You should be able to use it to check what you need for the horsemen.
Cheers 🍻

PR #147, opened 2026-03-29
feat: Add Adren HR Para Build & Update SoO

buddyleex:
- Added Adrenaline HR Paragon Build functions for setup, HR management, and rotation management
- Seemlessly tied in options into pre-existing Movement and Fight Functions
- Added Paragon class check to SoO bot. This check and different paragon builds can be added to all bots after testing.

### DIFF
--- a/lib/Build_PW_Heroic-Refrain.au3
+++ b/lib/Build_PW_Heroic-Refrain.au3
@@ -139,8 +139,10 @@ EndFunc
 ;~ Full scan every tick to avoid stale state from deaths, movement, or buff expiry.
 Func HRPhaseApplyParty()
 	Local Static $heroCount = GetHeroCount()
+	Local Static $nextHero = 1
 
-	For $i = 1 To $heroCount
+	For $j = 0 To $heroCount - 1
+		Local $i = Mod($nextHero + $j - 1, $heroCount) + 1
 		Local $agentID = GetHeroID($i)
 		If $agentID == 0 Then ContinueLoop
 
@@ -150,6 +152,7 @@ Func HRPhaseApplyParty()
 
 		If GetEffect($ID_HEROIC_REFRAIN, $i) == Null Then
 			UseSkillEx($BUILD_PW_HEROIC_REFRAIN, $agent)
+			$nextHero = Mod($i, $heroCount) + 1
 			Return $HR_PHASE_APPLY_PARTY
 		EndIf
 	Next


lib/Build_PW_Heroic-Refrain.au3
```diff
@@ -41,14 +41,355 @@ Global Const $BUILD_PW_BURNING_REFRAIN = 7
...
+		$hr_initialized = True
+	EndIf
+
+	; Always cast TOF on cooldown regardless of phase.
+	; TOF is off the global cooldown so it can be cast during HR casts, while knocked down, etc.
+	; Leadership returns ~10e from party members affected by shouts, making this energy-neutral.
+	; Under Quickening Zephyr, TOF recharges before its buff expires. Recasting while active
+	; does NOT trigger HR refresh, so we must wait for the old buff to expire first.
+	If IsRecharged($BUILD_PW_THEYRE_ON_FIRE) Then
+		Local $tofRemaining = GetEffectTimeRemaining($ID_THEYRE_ON_FIRE, 0)
```
caustic-kronos: - GetEffectTimeRemaining is expensive. That's why we often use timers instead.
Here it's useless, because HR para should be the only source of this shout, and the only thing that can make it still be on allies is if the duration of the shout was increased, which only happens with [Enduring Harmony](https://wiki.guildwars.com/wiki/Enduring_Harmony) or [Vocal Was Sogolon](https://wiki.guildwars.com/wiki/Vocal_Was_Sogolon).
Since this function is called every 10+ seconds, it will never happen that They're on fire is still active.

- There are possibilities for TOF to not be recasted, that's true. But in effect, this will happen only for [Energizing Wind](https://wiki.guildwars.com/wiki/Energizing_Wind), [Diversion](https://wiki.guildwars.com/wiki/Diversion) and [Edge of Reason](https://wiki.guildwars.com/wiki/Edge_of_Reason). Rare. I would drop the IsRecharged as well, the cost of checking recharge of shout is not worth the so rare cases when the skill won't be recharged.

lib/Build_PW_Heroic-Refrain.au3
```diff
@@ -41,14 +41,355 @@ Global Const $BUILD_PW_BURNING_REFRAIN = 7
...
+		Return
+	EndIf
+
+	$hr_phase = $HR_PHASE_MAINTENANCE
+	$hr_last_tof_time = TimerInit()
+EndFunc
+
+;~ Maintenance phase: reapply HR on rezzed heroes. TOF is handled at top of tick.
+Func HRPhaseMaintenance()
+	Local $playerAttr = GetMaxHRAttributeLevel(0)
```
caustic-kronos: Same as https://github.com/caustic-kronos/BotsHub/pull/147/changes#r3021354580

lib/Build_PW_Heroic-Refrain.au3
```diff
@@ -41,14 +41,355 @@ Global Const $BUILD_PW_BURNING_REFRAIN = 7
...
+
+	; Priority 6: "To the Limit!" (only if foes nearby for adrenaline gain)
+	If IsRecharged($BUILD_PW_TO_THE_LIMIT) And CountFoesInRangeOfAgent(GetMyAgent(), $RANGE_EARSHOT) > 0 Then
+		UseSkillEx($BUILD_PW_TO_THE_LIMIT)
+		Return
+	EndIf
+
+	; Priority 7: "For Great Justice!" (5e, lowest priority)
+	If IsRecharged($BUILD_PW_FOR_GREAT_JUSTICE) And GetEnergy() >= 5 Then
+		If GetEffect($ID_FOR_GREAT_JUSTICE, 0) == Null Then
```
caustic-kronos: Same comment for [For Great Justice](https://wiki.guildwars.com/wiki/%22For_Great_Justice!%22) - 20s duration for a 45s CD, there is no need to check that we don't have the effect anymore.

lib/Utils.au3
```diff
@@ -918,12 +918,17 @@ $default_move_aggro_kill_options.Add('openChests', True)
...
+$default_move_aggro_kill_options.Add('doWhileMoving', Null)
+$default_move_aggro_kill_options.Add('combatFunction', Null)
 ; default 60 seconds fight duration
 $default_move_aggro_kill_options.Add('fightDuration', 60000)
 
 Global $flag_move_aggro_kill_options = CloneDictMap($default_move_aggro_kill_options)
 $flag_move_aggro_kill_options.Item('flagHeroesOnFight') = True
 
+Global $active_fight_options = Null
+Global $active_flag_fight_options = Null
```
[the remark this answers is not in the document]
buddyleex: Resolved in bf36b84e5886546eb1fa7b0e132dd834e0991e1d

lib/Build_PW_Heroic-Refrain.au3
```diff
@@ -41,14 +41,278 @@ Global Const $BUILD_PW_BURNING_REFRAIN = 7
...
+; State resets automatically on death or zone change.
+; External reset is signalled via $hr_reset_pending; phase functions return
+; the next phase to keep phase state local to the tick function.
+; ============================================================================
+
+;~ AdlibRegister target. Performs at most one phase action per call.
+;~ TOF is cast unconditionally every tick to keep existing HR alive.
+Func TickHeroicRefrain()
+	Local Static $phase = $HR_PHASE_SELF_SETUP
+	Local Static $initialized = False
```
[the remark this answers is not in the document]
buddyleex: It's important to obtain +20 Leadership on the Paragon before buffing party but I see your point now...a binary check flow should implicitly initiate the double-cast needed to get the HR buffed from +3 to +4 breakpoint.

Tick 1: no HR → cast (player gets +3, 19 Leadership) → stay in self-setup
Tick 2: has HR → cast again (at 19 Leadership = +4) → advance to party phase

Heroes should only receive HR after self-setup completes so they always get the buffed +4 HR, and if it drops entirely due to death or energy drought, the binary check should catch and restart the two-cast sequence. HR can't partially degrade to a lower attribute level, it either exists or it doesn't.

Everything in 86fe11995c1e0f4332ea919a3b7306c1a059df56 should address this and the other concerns with that file.

caustic-kronos:
There is still the dictionary cloning that needs to be removed from MoveAggroAndKill, and I suggest going simple for now inside General Setup: check that we are Paragon and that we have HR, then we can AdlibRegister the maintenance function and replace the fight function with the paragon one. If not, nothing changes.

PR #151, opened 2026-04-04
fix: delay checking hero count

buddyleex:
Add a delay checking hero count so script doesn't make decisions based off erroneous counts in case heroes pre-exist in party.

### DIFF
--- a/lib/Utils.au3
+++ b/lib/Utils.au3
@@ -1571,6 +1571,18 @@ Func AddHeroByProfession($professionID, $preferredHeroID = 0)
 EndFunc
 
 
+;~ Try to add a specific hero to the party. No fallback to other heroes.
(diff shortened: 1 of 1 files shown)


lib/Utils.au3
```diff
@@ -1552,6 +1552,7 @@ EndFunc
 ;~ Iterates all known heroes of that profession and attempts AddHero until one succeeds.
 ;~ Returns the hero's party index (1-based) on success, or 0 if no hero of that profession could be added.
 Func AddHeroByProfession($professionID, $preferredHeroID = 0)
+	Sleep(500)
```
caustic-kronos: Why put a sleep here ?
The convention is to put the sleep after the operation that requires it. We do it lower - when we add hero, we sleep afterward. Doing it at beginning of the function: we lose the meaning of the sleep: what are we waiting for ?

PR #157, opened 2026-04-10
fix: Add pre-clear back to griffon run

buddyleex:
Add pre-clear back to griffon run, move take quest from wailing forest function into griffon run for more security.

Added a couple of "The Hunt" quest checks as it's a quest that gets missed sometimes for completion.


src/missions/FoW.au3
```diff
@@ -497,42 +497,77 @@ Func ForestOfTheWailingLord()
...
-	If Not IsPlayerOrPartyAlive() Then Return $FAIL
 	RandomSleep(5000)
 
 	Info('Kill Last Shard Wolf')
 	MoveAggroAndKill(-14430, -2750, '8')
 	MoveAggroAndKill(-18000, -3500, '9')
 	KillShardWolf()
 	MoveAggroAndKill(-17600, -4800, '10')
-	KillShardWolf()
+	KillShardWolf()
```
caustic-kronos: No need for spaces after function, let's revert those
```suggestion
	KillShardWolf()
```

src/missions/FoW.au3
```diff
@@ -497,42 +497,77 @@ Func ForestOfTheWailingLord()
...
-	RandomSleep(1000)
-	Info('Leading griffons back')
 	MoveAggroAndKill(-17300, 9600, '1')
 	MoveAggroAndKill(-16500, 8500, '2')
 	MoveAggroAndKill(-7500, 5000, '3')
 	MoveAggroAndKill(-6750, -4250, '4')
 	MoveAggroAndKill(-9500, -6000, '5')
 	MoveAggroAndKill(-13750, -2750, '6')
 	MoveAggroAndKill(-15750, -1750, '7')
-	If Not IsPlayerOrPartyAlive() Then Return $FAIL
```
caustic-kronos, line 519: We need to keep those failure tests at a somewhat regular rate - otherwise, if you die, you have to go through all the checks in all the following instructions until the end of the function.
That can take a few minutes - so it's better to have some checks, especially where the bot seems to die.

src/missions/FoW.au3
```diff
@@ -497,42 +497,77 @@ Func ForestOfTheWailingLord()
...
+		KillShardWolf()
+		MoveAggroAndKill(-17600, -4800, '10')
+		KillShardWolf() 
+		MoveAggroAndKill(-16575, -5200, '11')
+		KillShardWolf() 
+		MoveAggroAndKill(-18000, -3500)
+		MoveAggroAndKill(-13750, -2750)
+		CommandAll(-9800, -4800)
+		MoveAggroAndKill(-15750, -1750)
+		RandomSleep(5000)
```
caustic-kronos, line 572: EndIf missing here.

caustic-kronos (changes requested):
Concerning shardwolf, is it necessary to make checks at so many points ? Keep in mind that range to check shardwolf is 5000 (full compass range)

buddyleex:
so from my experience using fow script it does not always work for whatever reason. So I've added more moves and shardwolf checks to compensate. Many times I've gotten to the very end and it's missed 1 or even 2 shardwolves.

I'm not sure what is worng with my specific build or comptuer if it works completely fine for others?

src/missions/FoW.au3
caustic-kronos: Still valid

caustic-kronos:
Issue might come from AdlibRegister for HR.
AdlibRegister creates interruptions, for instance if you ran:
Move(x, y)
And you call an AdlibRegister that casts HR or does a Move(a, b), this last operation will overwrite the first move.
Same effect as if you just have them one after the other:
Move(x, y)
Move(a, b)
Your character will never end up in (x, y) because as soon as it starts going to (x, y) it receives the command to go to (a, b).

Anyway, we rarely use Move(), we use MoveTo usually, which doesn't have this issue, but my point is that some operations can be interrupted by other operations called afterward, like in an AdlibRegister call.
I suspect this might interrupt shardwolf chasing. I suspect it's the same thing that causes problems in SoO for @TDowg

src/missions/FoW.au3
buddyleex: I'm not sure why it's still showing it's an unresolved I added it back under MoveAggroAndKill(-15750, -1750, '7')

src/missions/FoW.au3
```diff
@@ -626,9 +680,13 @@ EndFunc
...
 Func KillShardWolf()
 	Local $foes = GetFoesInRangeOfAgent(GetMyAgent(), $RANGE_COMPASS, IsShardWolf)
-	If IsArray($foes) And UBound($foes) > 0 Then
+	While IsArray($foes) And UBound($foes) > 0
 		Local $shardWolf = $foes[0]
-		MoveAggroAndKill(DllStructGetData($shardWolf, 'X'), DllStructGetData($shardWolf, 'Y'))
-	EndIf
+		Local $wolfX = DllStructGetData($shardWolf, 'X')
+		Local $wolfY = DllStructGetData($shardWolf, 'Y')
+		MoveTo($wolfX, $wolfY)
```
caustic-kronos, line 687: MoveTo is dangerous - if there are still ennemies in the way character will just move to the position of the wolf without making any attack. We don't need it, let's remove.
Otherwise all good

caustic-kronos (approved):

caustic-kronos (approved):

PR #159, opened 2026-04-19
feat: Add a checkbox to store items that should be kept.

Azgw:
Currently when weapons and armors have mods that we want to keep, the selling/storing logic would just keep them in inventory. This means that the bot will progressively start runs with a full inventory. This offers to store the items that contains a mod that should be kept and add a checkbox to the GUI. A better solution would be to extract the mod and store it, but I don't know how to do it.

### DIFF
--- a/BotsHub.au3
+++ b/BotsHub.au3
@@ -130,6 +130,7 @@ $run_options_cache['run.use_consets'] = False
 $run_options_cache['run.use_scrolls'] = False
 $run_options_cache['run.sort_items'] = False
 $run_options_cache['run.farm_materials_mid_run'] = False
+$run_options_cache['run.store_items_to_keep'] = False
 $run_options_cache['run.bags_count'] = 5
 $run_options_cache['run.donate_faction_points'] = True
 $run_options_cache['run.buy_faction_scrolls'] = False
@@ -404,6 +405,7 @@ Func ReadConfigFromJson($jsonString)
 	$run_options_cache['run.loop_mode'] = _JSON_Get($jsonObject, 'run.loop_mode')
 	$run_options_cache['run.hard_mode'] = _JSON_Get($jsonObject, 'run.hard_mode')
 	$run_options_cache['run.farm_materials_mid_run'] = _JSON_Get($jsonObject, 'run.farm_materials_mid_run')
+	$run_options_cache['run.store_items_to_keep'] = _JSON_Get($jsonObject, 'run.store_items_to_keep')
 	$run_options_cache['run.consume_consumables'] = _JSON_Get($jsonObject, 'run.consume_consumables')
 	$run_options_cache['run.use_consets'] = _JSON_Get($jsonObject, 'run.use_consets')
 	$run_options_cache['run.use_scrolls'] = _JSON_Get($jsonObject, 'run.use_scrolls')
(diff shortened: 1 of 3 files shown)


caustic-kronos:
Hello !
Thanks for PR and sorry for latency, I am not at home and looking at code on a phone is really tedious.

Anyway, I looked at PR but I'm curious about the premise of this PR. There are already checkboxes to store items that are kept in the loot tab.

Typically, if you want to store items with the mods you want, you just tick the store items.weapons.gold (or something like that I don't have the GUI to check exact path).
The bot will keep the items with the mods you chose. Then once it goes into the storing function, it will store the chosen items from the interface.

So there shouldn't be a need to add anything. Only to tick those boxes.

Unless I am missing something ?

Azgw:
Hello,

Thanks for the review. Maybe I misunderstood how some things are working, but in the inventory tab, the option to store golds items (weapons and armors) are based on the type of the item, its rarity and the requirement. For instance, a gold Q9 wand.
I think the bot then stores all weapons that meet the requirement (so all Q9 gold wands), regardless of the weapon components chosen from in "Keep components".

So, if I select some components that I want to keep but not select items to store, then the items stay in the inventory. And If I select items to store, then too many items get stored.

If I misunderstood the expected behavior, can you explain how I should proceed?

caustic-kronos:
Hello !
I'm back from hiatus (family health issues) and finally have some access to my computer.

You didn't misunderstand this behaviour - but maybe the idea behind how the bot deals with inventory.
It operates by steps:
- identifies
- salvages
- sells
- stores

The components choice impacts the salvage and sell steps - we don't salvage or sell items with good/chosen components.
So normally, if you decide you want to keep some components, the items containing those won't be sold/salvaged, so they will be left in the inventory at the end of those steps.
Then if you tick store all Q9 gold wands, you will indeed store them all. But normally, the "bad" wands, the one with no good components, no good skins and such, should be sold or salvaged. So at the store steps, only the good ones that are either rare or have the components you chose should be stored.
So normally you should never have "too many items get stored" if you configured the bot to properly sell/salvage the bad items.

Hope this is clear, and if you think that this way of doing things is not good enough, don't hesitate to explain more in details

Azgw:
Hey,

I hope everything is alright !

I looked into the code in more details, but just to confirm. Say I want to keep items that have valuable components, and put them in storage, but sell all other items. The idea is to maximize item space in the bag, then I should 1/ properly tick the components that I want to keep, 2/ select everything as items to sell and 3/ select everything as items to store? Then the bot tries to everything but items with components to keep are not sold then then they are stored?

It is currently not so obvious looking only from the GUI. Thanks a bunch !

caustic-kronos:
Everything not alright but life goes on 🥲 

You're correct.
If you select everything to sell, some components to keep, and everything to store, then the bot will:
- sell everything
- except the items containing your desired components
- and except some specific items (rare items like glacial blades, or low req items like q8 11-22 sword)
- then it will store everything (that includes the aforementioned items containing desired components, the glacial blade and the q8 11-22 sword)

That's why currently, in the default configuration, all weapons are set to be sold.

Azgw:
Hi again,

I wish you all the best then !

Alright, then I did not understand the behavior and my PR is useless  :)

Cheers,

PR #161, opened 2026-04-23
fix:		Fixed gold balancing logic

Azgw:
With the previous logic, we would not deposit gold to the bank if we had no gold in the bank or not withdraw gold when there was too puch gold in the bank.

### DIFF
--- a/lib/Utils-Storage.au3
+++ b/lib/Utils-Storage.au3
@@ -1502,18 +1502,28 @@ Func BalanceCharacterGold($goldAmount, $mode = 0)
 	Info('Balancing characters gold')
 	Local $goldCharacter = GetGoldCharacter()
 	Local $goldStorage = GetGoldStorage()
-	If $goldStorage > 950000 Then
-		Warn('Too much moneyz in Xunlai, you are a rich bitch.')
-	ElseIf $goldStorage < 50000 Then
-		Warn('Not enough moneyz in Xunlai, you poor bastard.')
-	ElseIf $goldCharacter > $goldAmount And $mode <> 1 Then
-		DepositGold($goldCharacter - $goldAmount)
-		Info('Deposited gold')
+	If $goldCharacter > $goldAmount And $mode <> 1 Then
+		; We want to deposit
+		If $goldStorage > (1000000 - ($goldCharacter - $goldAmount)) Then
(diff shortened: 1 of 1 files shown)


lib/Utils-Storage.au3
```diff
@@ -1502,18 +1502,28 @@ Func BalanceCharacterGold($goldAmount, $mode = 0)
...
-	If $goldStorage > 950000 Then
-		Warn('Too much moneyz in Xunlai, you are a rich bitch.')
-	ElseIf $goldStorage < 50000 Then
-		Warn('Not enough moneyz in Xunlai, you poor bastard.')
-	ElseIf $goldCharacter > $goldAmount And $mode <> 1 Then
-		DepositGold($goldCharacter - $goldAmount)
-		Info('Deposited gold')
+	If $goldCharacter > $goldAmount And $mode <> 1 Then
+		; We want to deposit
+		If $goldStorage > (1000000 - ($goldCharacter - $goldAmount)) Then
```
caustic-kronos:
```suggestion
		Local $depositAmount = $goldCharacter - $goldAmount
		If $goldStorage > (1000000 - $depositAmount) Then
```
Let's refactor $goldCharacter - $goldAmount into a variable to make it clearer what it is (and replace the same value at line 1510)

lib/Utils-Storage.au3
```diff
@@ -1502,18 +1502,28 @@ Func BalanceCharacterGold($goldAmount, $mode = 0)
...
+		Else
+			DepositGold($goldCharacter - $goldAmount)
+			Info('Deposited gold')
+		EndIf
+		Return True
 	ElseIf $goldCharacter < $goldAmount And $mode <> 2 Then
-		WithdrawGold($goldAmount - $goldCharacter)
-		Info('Withdrawn gold')
+		; We want to withdraw
+		If $goldStorage < ($goldAmount - $goldCharacter) Then
```
caustic-kronos: We can do the same thing here

caustic-kronos (changes requested):
Good fix, one small addition and it will be clean

Azgw:
Did the change you asked for !

caustic-kronos (approved):

caustic-kronos:
Thank you ! Let's merge 👍

PR #163, opened 2026-04-26
feat:	Add 'Open Storage' button to GUI

buddyleex:
Added Open Storage button to GUI that calls to initialize a command structure derived from GWToolBox through the CommandUIMsg assember path (kShowXunlaiChest) wParam layout:
    - h0000 = 0
    - storage_pane_unlocked = 1
    - anniversary_pane_unlocked = 1

### DIFF
--- a/lib/BotsHub-GUI.au3
+++ b/lib/BotsHub-GUI.au3
@@ -102,7 +102,7 @@ Global $gui_group_teamoptions, $gui_teamlabel, $gui_teammemberlabel, $gui_teamme
 		$gui_checkbox_load_build_hero_4, $gui_checkbox_load_build_hero_5, $gui_checkbox_load_build_hero_6, $gui_checkbox_load_build_hero_7, _
 		$gui_label_build_hero_1, $gui_label_build_hero_2, $gui_label_build_hero_3, $gui_label_build_hero_4, $gui_label_build_hero_5, $gui_label_build_hero_6, $gui_label_build_hero_7, _
 		$gui_input_build_player, $gui_input_build_hero_1, $gui_input_build_hero_2, $gui_input_build_hero_3, $gui_input_build_hero_4, $gui_input_build_hero_5, $gui_input_build_hero_6, $gui_input_build_hero_7
-Global $gui_group_otheroptions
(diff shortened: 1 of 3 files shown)


caustic-kronos:
That's interesting and a very good addition.
I need to test it a bit, will validate once I have access to my puter

buddyleex:
It works (mostly). For  some reason I can't get the anniversary tab to work/show. And I'm not sure what Copilot is talking about.

The assembly shows exactly what happens:

```asm
CommandUIMsg:
push 0
mov edx, eax
add edx, 8         ; edx = pointer to data at struct offset +8
push edx           ; UIMessage 2nd arg = POINTER to wParam struct
push dword[eax+4]  ; UIMessage 1st arg = msgID
call UIMessage
```

UIMessage receives a **pointer** to the wParam data at `eax+8` — it reads the struct from memory, not a packed dword. Copilot confused "pointer to payload" with "payload is a single dword."

The GW struct for `kShowXunlaiChest` is:
```
uint32_t h0000 = 0          → bytes 0–3  at offset +8
bool storage_pane_unlocked  → byte 4     at offset +12
bool anniversary_pane_unlocked → byte 5  at offset +13
```

Our `'ptr;dword;dword;byte;byte'` on 32-bit GW maps to:
| Field | Type | Offset | Value |
|-------|------|--------|-------|
| 1 | ptr | +0 | CommandUIMsg fn ptr |
| 2 | dword | +4 | 0x10000040 (msgID) |
| 3 | dword | +8 | 0 (h0000) |
| 4 | byte | +12 | 1 (storage_pane_unlocked) |
| 5 | byte | +13 | 1 (anniversary_pane_unlocked) |

This matches the GWCA struct exactly. Copilot's `0x01010000` suggestion would pack the flag bytes into the upper half of the h0000 dword (offsets +10/+11), which is wrong. **No change needed.**

caustic-kronos:
Tested it - it seems to work fine.
What do you mean by "I can't get the anniversary tab to work/show" ?
I'll merge - main functionality is there and works nicely, we can always iterate and add/fix small issues.

caustic-kronos (approved):

PR #165, opened 2026-04-28
Update memory address in AssemblerCreateScanProcedure

buddyleex:
Fix scanner range to update merchant functions due to new build.


caustic-kronos (approved):
Thanks for the fix - I'm not at home for 2 weeks already and counting so it helps 👍 
Merging ...

PR #168, opened 2026-04-30
fix: Update CommandTradeInitiate

buddyleex:
The GW patch changed the UIMessage calling convention for trade initiation. CommandTradeInitiate now pushes an extra 0 argument and stack cleanup is updated from 8 to 12 bytes accordingly. If we ever create a UIMSG_INITIATE_TRADE variable make sure we use new address 0x100001B3.


caustic-kronos (approved):
Thank you for fix - let's merge

PR #170, opened 2026-05-04
fix: Correct AddHeroByProfession logic

buddyleex:
Would fail when hero already persisted in party add checks for hero in function to make sure to proceed without relying on LeaveParty() as a bandaid.

### DIFF
--- a/lib/Utils.au3
+++ b/lib/Utils.au3
@@ -1618,13 +1618,29 @@ EndFunc
 ;~ Try to add any available hero of the given profession to the party.
 ;~ If $preferredHeroID is specified, tries that hero first before falling back to others.
 ;~ Iterates all known heroes of that profession and attempts AddHero until one succeeds.
+;~ If a hero of that profession is already in party, returns that hero number and does not try to add.
 ;~ Returns the hero's party index (1-based) on success, or 0 if no hero of that profession could be added.
 Func AddHeroByProfession($professionID, $preferredHeroID = 0)
+	If $preferredHeroID > 0 Then
+		Local $preferredHeroNumber = GetHeroNumberByHeroID($preferredHeroID)
+		If IsNumber($preferredHeroNumber) Then Return $preferredHeroNumber
+	EndIf
+
+	For $heroNumber = 1 To GetHeroCount()
+		If GetHeroProfession($heroNumber) == $professionID Then Return $heroNumber
+	Next
+
 	If $preferredHeroID > 0 Then
 		Local $previousCount = GetHeroCount()
 		AddHero($preferredHeroID)
 		Sleep(500)
-		If GetHeroCount() > $previousCount Then Return $SUCCESS
+		If GetHeroCount() > $previousCount Then
+			Local $preferredHeroNumber = GetHeroNumberByHeroID($preferredHeroID)
+			If IsNumber($preferredHeroNumber) Then Return $preferredHeroNumber
+			For $heroNumber = 1 To GetHeroCount()
+				If GetHeroProfession($heroNumber) == $professionID Then Return $heroNumber
+			Next
+		EndIf
 	EndIf
 	For $heroID In MapKeys($HERO_PROFESSIONS)
 		If $HERO_PROFESSIONS[$heroID] <> $professionID Then ContinueLoop
@@ -1632,7 +1648,13 @@ Func AddHeroByProfession($professionID, $preferredHeroID = 0)
 		Local $previousCount = GetHeroCount()
 		AddHero($heroID)
 		Sleep(500)
-		If GetHeroCount() > $previousCount Then Return $SUCCESS
+		If GetHeroCount() > $previousCount Then
+			Local $heroNumber = GetHeroNumberByHeroID($heroID)
+			If IsNumber($heroNumber) Then Return $heroNumber
+			For $i = 1 To GetHeroCount()
+				If GetHeroProfession($i) == $professionID Then Return $i
+			Next
+		EndIf
 	Next
 	Error('Could not add any hero with profession ID ' & $professionID)
 	Return $FAIL
--- a/lib/Utils.au3
+++ b/lib/Utils.au3
@@ -1619,7 +1619,7 @@ EndFunc
 ;~ If $preferredHeroID is specified, tries that hero first before falling back to others.
 ;~ Iterates all known heroes of that profession and attempts AddHero until one succeeds.
 ;~ If a hero of that profession is already in party, returns that hero number and does not try to add.
-;~ Returns the hero's party index (1-based) on success, or 0 if no hero of that profession could be added.
+;~ Returns the hero's party index (1-based) on success, or Null if no hero of that profession could be added.
 Func AddHeroByProfession($professionID, $preferredHeroID = 0)
 	If $preferredHeroID > 0 Then
 		Local $preferredHeroNumber = GetHeroNumberByHeroID($preferredHeroID)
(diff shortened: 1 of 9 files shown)


src/farms/Raptors.au3
```diff
@@ -95,6 +95,7 @@ $raptors_move_options.Item('openChests')				= False
 
 Global $raptors_farm_setup = False
 Global $raptors_player_profession = $ID_WARRIOR
+Global $raptors_hero_number = 1
```
caustic-kronos, line 98: I moved away from having the hero number in those farms that use very few heroes (1 to 3 usually) because it feels like it doesn't really provide valuable information.

I think adding $raptors_hero_number everywhere here doesn't really make code clearer - since it's a single hero, it doesn't really clear a misunderstanding, and having 1 makes code shorter as well.

That's up for debate though, to be honest. What do you think ?

src/farms/Raptors.au3
```diff
@@ -150,11 +151,15 @@ Func SetupTeamRaptorsFarm()
 	Info('Setting up team')
 	RandomSleep(500)
 	LeaveParty()
-	If AddHeroByProfession($ID_PARAGON, $RAPTORS_HERO_PARTY_ID) == $FAIL Then Return $FAIL
+	$raptors_hero_number = AddHeroByProfession($ID_PARAGON, $RAPTORS_HERO_PARTY_ID)
+	If Not IsNumber($raptors_hero_number) Then
+		Warn('Could not add paragon hero to party')
+		Return $FAIL
+	EndIf
```
caustic-kronos, line 158: I think this is a clear downgrade.
We are encoding the function inner logic (hero was or was not properly added) outside of the function, which means that we need that check absolutely everywhere.
We should revert this and keep AddHeroByProfession on a boolean logic - or if you want we can keep AddHeroByProfession returning hero number but then we add an additional function layer checking return of it against expected hero number and use that one everywhere.
Personally I find that second option overkill and it adds complexity for no real gains, I would revert to previous behaviour.

---

Additional question: why get hero number from the function ? We only use a single hero anyway, so its number should always be 1. Which would also make raptors_hero_number a Global Const instead of a global and we could test against it ?

src/farms/GemstoneMargonite.au3
```diff
@@ -189,8 +189,13 @@ Func SetupTeamMargoniteFarm()
 	If IsTeamAutoSetup() Then Return $SUCCESS
 
 	Info('Setting up team')
+	Local $monkHeroNumber = Null
 	LeaveParty()
-	If AddHeroByProfession($ID_MONK, $MARGONITE_HERO_PARTY_ID) == $FAIL Then Return $FAIL
+	$monkHeroNumber = AddHeroByProfession($ID_MONK, $MARGONITE_HERO_PARTY_ID)
+	If Not IsNumber($monkHeroNumber) Then
+		Warn('Could not add monk hero to party')
+		Return $FAIL
```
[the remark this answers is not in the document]
caustic-kronos: Good job buddy ! 😆

lib/Utils.au3
```diff
@@ -1618,24 +1618,46 @@ EndFunc
 ;~ Try to add any available hero of the given profession to the party.
 ;~ If $preferredHeroID is specified, tries that hero first before falling back to others.
 ;~ Iterates all known heroes of that profession and attempts AddHero until one succeeds.
-;~ Returns the hero's party index (1-based) on success, or 0 if no hero of that profession could be added.
+;~ If a hero of that profession is already in party, returns that hero number and does not try to add.
+;~ Returns the hero's party index (1-based) on success, or Null if no hero of that profession could be added.
 Func AddHeroByProfession($professionID, $preferredHeroID = 0)
```
caustic-kronos, line 1623: We need to warn in description of this function that we can't use it twice with the same profession.

lib/Utils.au3
```diff
@@ -1618,24 +1618,46 @@ EndFunc
...
+	If $preferredHeroID > 0 Then
+		Local $preferredHeroNumber = GetHeroNumberByHeroID($preferredHeroID)
+		If IsNumber($preferredHeroNumber) Then Return $preferredHeroNumber
+	EndIf
```
caustic-kronos, line 1627: We should instead move this to the existing `If $preferredHeroID > 0` condition

lib/Utils.au3
```diff
@@ -1618,24 +1618,46 @@ EndFunc
...
 		AddHero($preferredHeroID)
 		Sleep(500)
-		If GetHeroCount() > $previousCount Then Return $SUCCESS
+		If GetHeroCount() > $previousCount Then
+			Local $preferredHeroNumber = GetHeroNumberByHeroID($preferredHeroID)
+			If IsNumber($preferredHeroNumber) Then Return $preferredHeroNumber
+			For $heroNumber = 1 To GetHeroCount()
+				If GetHeroProfession($heroNumber) == $professionID Then Return $heroNumber
+			Next
+		EndIf
```
caustic-kronos, line 1643: This is confusing:
1- We check if preferred hero is present, if so we return its number
2- We check if hero of given profession is present, is so we return its number
3- We try adding preferred hero
4- If we added a hero:
a- we get its number and return it if we got it (we should have it since we added successfully a hero so the test is not required
b- we loop over heroes of the given profession and return a hero with the given profession (but we already successfully added a hero so why is this required ?)

Overall, the logic in this function is too complicated and error prone.
It also doesn't really do what we want it to do anymore:
Calling AddHeroByProfession, we expect we will add a hero of given profession, we don't really care about what is already in the team.

caustic-kronos (commented):
Hello !
While I see what you tried to solve here, I think there are several design issues.
Mainly:
- the contract of AddHeroByProfession is now blurry. Its name implies we add a hero by the given profession with a potential preference for a specific hero. It has nothing to do with what is already present in the group. I think changing this function to give it knowledge of what is in the group is not the proper way to go about it.
- in particular, this makes it impossible to add several heroes of the same profession to the group with this function
- hero number should be there to make code clearer, but in the case of very small amount of heroes I'm not sure it succeeds in this endeavour. This is debatable though.
- regardless of whether we keep hero number or not, returning it by AddHeroByProfession and comparing it in the bots themselves is moving that logic in the wrong place. It should not be up to the bots themselves to do this whole testing logic:
```
$corsairs_dunkoro_hero_number = GetHeroNumberByHeroID($ID_DUNKORO)
	If Not IsNumber($corsairs_dunkoro_hero_number) Then
		Warn('Could not find Dunkoro in party')
		Return $FAIL
	EndIf
```
- I'm not sure what's the purpose of returning hero position either considering we have a fixed amount of heroes anyway - which means that the Global variables could stay Global Const

caustic-kronos (changes requested):

buddyleex:
I'm seriously considering closing this PR, the problem this PR was meant to solve was LeaveParty() was not working. If we start with a fresh slate, the current functions work fine. So instead of fixing the root cause I came up with this convoluted mess.

Updating the correct parameter for Kick All Heroes should correct the issues we have seen since the updates:

;~ Kicks all heroes from the party.
Func KickAllHeroes()
	Return SendPacket(0x8, $HEADER_HERO_KICK, 0x28)
EndFunc

buddyleex:
I think we can close this I tested the fix for KickAllHeroes with the old function behavior and it works as it should. Addressed in 7026778840c5a257967f92ed54cf700e35798a5d

caustic-kronos:
Yes, I think it avoids making things more complicated than they need to be. I'll close this one and merge the other.
Thank you for looking into this and fixing the issue 👍

PR #173, opened 2026-05-15
feat: harden Follower with leash range and outpost leader resolution

### LINKED ISSUE #152: Follower Bot
Not sure if I am doing something wrong, but my group is following the 2nd person in the party and not the first. I turn them all on once outside of outpost

marioweid:
- Add FOLLOWER_LEASH_RANGE (850) so the bot only engages/loots while the leader is within range, and breaks out of combat once the leader walks past it.
- Auto-resign once per explorable entry; reset state and wait for the map to finish loading on every map change; skip cinematics.
- Add FollowerResolveLeader, which reads the player record array via MemoryReadPtr and looks up the leader's agent by LoginNumber. The library's GetFirstPlayerOfParty returns nothing in outposts because party agent structs report LoginNumber=0 there.
- Rewrite GetFirstPlayerOfParty to use the same memory walk and match the result back to a party agent struct.

closes #152

### DIFF
--- a/lib/GWA2.au3
+++ b/lib/GWA2.au3
@@ -1160,6 +1160,10 @@ EndFunc
 
 
 #Region Party
+;~ Shared pointer chains into party/player structures. Tails are appended to reach sub-fields.
+Global Const $PARTY_CONTEXT_OFFSET[] = [0, 0x18, 0x4C, 0x54]
+Global Const $PLAYER_AGENT_ARRAY_OFFSET[] = [0, 0x18, 0x2C, 0x80C]
+
 ;~	Description: Returns different States about Party. Check with BitAND.
 ;~	0x8 = Leader starts Mission / Leader is travelling with Party
 ;~	0x10 = Hardmode enabled
@@ -1208,6 +1212,30 @@ EndFunc
 
 Func GetPartyWaitingForMission()
 	Return GetPartyState(0x8)
+EndFunc
+
+
+;~ Returns the number of player slots in the party player array (includes empty slots).
+Func GetPlayerCount()
+	Local $offset[] = [$PARTY_CONTEXT_OFFSET[0], $PARTY_CONTEXT_OFFSET[1], $PARTY_CONTEXT_OFFSET[2], $PARTY_CONTEXT_OFFSET[3], 0xC]
+	Local $result = MemoryReadPtr(GetProcessHandle(), $base_address_ptr, $offset)
+	Return $result[1]
+EndFunc
+
+
+;~ Returns the login number of the player at the given party slot index. Empty slots return 0.
+Func GetPartyPlayerLoginNumber($index)
+	Local $offset[] = [$PARTY_CONTEXT_OFFSET[0], $PARTY_CONTEXT_OFFSET[1], $PARTY_CONTEXT_OFFSET[2], $PARTY_CONTEXT_OFFSET[3], 0x4, 4 * $index]
+	Local $result = MemoryReadPtr(GetProcessHandle(), $base_address_ptr, $offset)
+	Return $result[1]
+EndFunc
+
+
+;~ Returns the agent ID for a given login number by walking the player record array (80-byte records).
+Func GetAgentIDByLoginNumber($loginNumber)
+	Local $offset[] = [$PLAYER_AGENT_ARRAY_OFFSET[0], $PLAYER_AGENT_ARRAY_OFFSET[1], $PLAYER_AGENT_ARRAY_OFFSET[2], $PLAYER_AGENT_ARRAY_OFFSET[3], 80 * $loginNumber]
+	Local $result = MemoryReadPtr(GetProcessHandle(), $base_address_ptr, $offset)
+	Return $result[1]
 EndFunc
 #EndRegion Party
 
--- a/src/utilities/Follower.au3
+++ b/src/utilities/Follower.au3
@@ -37,6 +37,8 @@ Global Const $FOLLOWER_INFORMATIONS = 'This bot makes your character follow the 
 	& 'It will loot all items it can loot.' & @CRLF _
 	& 'It will also loot all chests in range.'
 
+Global Const $FOLLOWER_LEASH_RANGE = 850
+
 ; Skill numbers declared to make the code WAY more readable (UseSkillEx($RAPTORS_MARK_OF_PAIN) is better than UseSkillEx(1))
 Global $player_profession_ID
 Global $follower_attack_skill_1 = Null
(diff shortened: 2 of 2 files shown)


src/utilities/Follower.au3
```diff
@@ -318,18 +353,60 @@ Func ParagonFight()
...
+;~ Resolve the leader's agent struct by reading the agent ID directly out of the player record array.
+;~ Works in both outposts and explorables. Bypasses the lib's GetFirstPlayerOfParty, which fails in
+;~ outposts because party agent structs report LoginNumber=0 there.
+;~ Player records are 80 bytes wide; agent ID is at offset 0 of each record.
+Func FollowerResolveLeader()
+	Local $processHandle = GetProcessHandle()
+	Local $selfLoginNumber = DllStructGetData(GetMyAgent(), 'LoginNumber')
+
+	Local $countOffset[] = [0, 0x18, 0x4C, 0x54, 0xC]
+	Local $playerCount = MemoryReadPtr($processHandle, $base_address_ptr, $countOffset)[1]
```
caustic-kronos: Let's move everything that has to do with raw memory read to GWA2 file - bots shouldn't have this responsibility.
We can create a function GetPlayerCount with this, and other related functions, and call them here.

Additionally, some of those offsets are used in other places (this one is in a function below and in GetPartySize). It would make sense to regroup them so that when there are updates we don't have to change them in many places.

src/utilities/Follower.au3
```diff
@@ -318,18 +353,60 @@ Func ParagonFight()
...
+
+		Return GetAgentByID($leaderAgentID)
 	Next
+
 	Return Null
+EndFunc
+
+
+;~ Get first player of the party team other than yourself. If no other player found in the party team then function returns Null
+Func GetFirstPlayerOfParty()
```
caustic-kronos, line 371: This function is now unused, so we could remove it, but I think keeping it could be handy.
Though it needs to be a little bit reverted to fit the same comments I made higher: we shouldn't make any raw memory read in here, all of that is done in GWA2 library files. Let's extract the memory read parts into functions we can use here.

src/utilities/Follower.au3
```diff
@@ -132,31 +134,64 @@ EndFunc
 
 ;~ Follower loop
 Func FollowerLoop($runFunction = DefaultRun, $fightFunction = DefaultFight)
-	If GetMapType() <> $ID_EXPLORABLE Then
-		Sleep(3000)
-		Return
+	Local Static $firstPlayer = Null, $currentMap = Null, $resignedThisMap = False
+
+	SkipCinematic()
```
caustic-kronos: I'm a bit worried about calling SkipCinematic constantly here, since FollowerLoop is called in a loop, we will try to skip cinematic every few seconds sometimes.
I did not look into it, but it would be great if we could only skip cinematic when actually in cinematic. Is there a way to detect that ?

src/utilities/Follower.au3
```diff
@@ -132,31 +134,64 @@ EndFunc
...
+
+	Local $mapID = GetMapID()
+	If $mapID <> $currentMap Then
+		$currentMap = $mapID
+		$firstPlayer = Null
+		$resignedThisMap = False
```
caustic-kronos: We can probably keep this as just "$resigned", no ? It's kind of a given that it's on a "per map" basis

src/utilities/Follower.au3
```diff
@@ -132,31 +134,64 @@ EndFunc
...
+		EndIf
+
+		Local $leaderID = DllStructGetData($firstPlayer, 'ID')
+		Local $me = GetMyAgent()
+		Local $leaderClose = False
+		If GetAgentExists($leaderID) Then
+			$leaderClose = (GetDistance($me, GetAgentByID($leaderID)) <= $FOLLOWER_LEASH_RANGE)
+		EndIf
+
+		If $leaderClose Then
```
caustic-kronos: $leaderClose is redundant and could be removed by changing code like so:

```suggestion
		If GetAgentExists($leaderID) And (GetDistance($me, GetAgentByID($leaderID)) <= $FOLLOWER_LEASH_RANGE) Then
```

src/utilities/Follower.au3
```diff
@@ -132,31 +134,64 @@ EndFunc
...
+			Local $foesCount = CountFoesInRangeOfAgent($me, $RANGE_EARSHOT)
+			If $foesCount > 0 Then
```
caustic-kronos: If clause is redundant with While clause, but I understand you added it for debug reason.
If you still need that debug we can keep it, otherwise we can remove the If and the two Debug calls.

caustic-kronos (changes requested):
Hello !
Thanks for the PR. Good additions I believe, this Follower bot really needed some love.

I made some comments, just a few corrections of logic here and there and two worries about skipping Cinematic too frequently and having memory reading logic inside Follower (we should keep that in GWA library files).

src/utilities/Follower.au3
marioweid: Hey sure, i didn't want to modify those files since they could be used by other users or different scripts.
Will look into it after my vocation.

src/utilities/Follower.au3
marioweid: Good point, need to check.

src/utilities/Follower.au3
marioweid: SkipCinematic now fires once on map entry, this helps for cinematics with map change, f.e. end of mission but not with cinematics without map change, havent found a solution for that

src/utilities/Follower.au3
marioweid: memory reads moved to GWA2 wrappers (GetPlayerCount, GetPartyPlayerLoginNumber, GetAgentIDByLoginNumber) with shared offset consts;

src/utilities/Follower.au3
marioweid: SkipCinematic now fires once on map entry;

src/utilities/Follower.au3
marioweid: $resignedThisMap → $resigned;

src/utilities/Follower.au3
marioweid: - dropped $leaderClose 
- the redundant debug calls
- inlined the leash check

src/utilities/Follower.au3
caustic-kronos: Good enough if we find a better solution that can still be another PR 👍

lib/GWA2.au3
```diff
@@ -1160,6 +1160,10 @@ EndFunc
 
 
 #Region Party
+;~ Shared pointer chains into party/player structures. Tails are appended to reach sub-fields.
+Global Const $PARTY_CONTEXT_OFFSET[] = [0, 0x18, 0x4C, 0x54]
+Global Const $PLAYER_AGENT_ARRAY_OFFSET[] = [0, 0x18, 0x2C, 0x80C]
```
caustic-kronos, line 1165: This is not great, but I don't have a better solution right now, so let's keep as is.

caustic-kronos (approved):

caustic-kronos:
Looking great now !
I think we could do better for the offset array, but that's an issue for another day, right now it looks good, let's merge 😄 
Thanks !

PR #174, opened 2026-05-18
Add go-offline and whisper-flash features; refactor ChatLog polling

buddyleex:
- add .files and .folders/ to gitignore except .github/ and .gitignore
- Add 'go offline' option: sets player status to offline when bot starts
- Add 'flash whisper' option: flashes GW taskbar button on incoming whisper
- Add corresponding GUI checkboxes with tooltips
- Integrate options into config JSON serialization/deserialization
- Refactor ChatLog detour from PostMessage/GUIRegisterMsg to adlib polling (1s atm)
  - Fixes stale hook detection and graceful cleanup on exit
  - Filters out zone-transition GW-encoded messages from real whispers
  - Improves reliability and eliminates cross-thread API issues
- Remove undefined GetPlayerStatus() call from SetPlayerStatus validation (can add back if we figure out proper function, gwau3 is missing too)

### DIFF
--- a/BotsHub.au3
+++ b/BotsHub.au3
@@ -204,8 +204,6 @@ Func Main()
 		EndIf
 		; Authentication
 		Authentification($character_name)
-		If $run_options_cache['run.go_offline'] Then SetPlayerStatus(0)
-		If $run_options_cache['run.flash_whisper'] Then EnableWhisperFlash()
 		$runtime_status = 'RUNNING'
 	Else
 		MsgBox(0, 'Error', 'Unknown run mode: ' & $run_mode)
@@ -416,6 +414,7 @@ Func ReadConfigFromJson($jsonString)
 	$run_options_cache['run.sort_items'] = _JSON_Get($jsonObject, 'run.sort_items')
 	$run_options_cache['run.collect_data'] = _JSON_Get($jsonObject, 'run.collect_data')
 	$run_options_cache['run.go_offline'] = _JSON_Get($jsonObject, 'run.go_offline')
+	$run_options_cache['run.flash_whisper'] = _JSON_Get($jsonObject, 'run.flash_whisper')
 	$run_options_cache['run.donate_faction_points'] = _JSON_Get($jsonObject, 'run.donate_faction_points')
 	$run_options_cache['run.buy_faction_resources'] = _JSON_Get($jsonObject, 'run.buy_faction_resources')
 	$run_options_cache['run.buy_faction_scrolls'] = _JSON_Get($jsonObject, 'run.buy_faction_scrolls')
@@ -464,6 +463,7 @@ Func WriteConfigToJson()
 	_JSON_addChangeDelete($jsonObject, 'run.sort_items', $run_options_cache['run.sort_items'])
 	_JSON_addChangeDelete($jsonObject, 'run.collect_data', $run_options_cache['run.collect_data'])
 	_JSON_addChangeDelete($jsonObject, 'run.go_offline', $run_options_cache['run.go_offline'])
+	_JSON_addChangeDelete($jsonObject, 'run.flash_whisper', $run_options_cache['run.flash_whisper'])
 	_JSON_addChangeDelete($jsonObject, 'run.donate_faction_points', $run_options_cache['run.donate_faction_points'])
 	_JSON_addChangeDelete($jsonObject, 'run.buy_faction_resources', $run_options_cache['run.buy_faction_resources'])
 	_JSON_addChangeDelete($jsonObject, 'run.buy_faction_scrolls', $run_options_cache['run.buy_faction_scrolls'])
@@ -608,6 +608,10 @@ Func GeneralFarmSetup()
 		If GetMapType() <> $ID_OUTPOST Then TravelToOutpost($ID_EYE_OF_THE_NORTH)
 		SetupPlayerUsingGlobalSettings()
 		SetupTeamUsingGlobalSettings()
+	EndIf
+	If $character_name <> '' Then
+		If $run_options_cache['run.go_offline'] Then SetPlayerStatus(0)
+		If $run_options_cache['run.flash_whisper'] Then EnableWhisperFlash()
 	EndIf
 
 	SetupPlayerBuildOverrides()
--- a/BotsHub.au3
+++ b/BotsHub.au3
@@ -33,6 +33,7 @@
 #include 'lib/GWA2_ID.au3'
 #include 'lib/GWA2.au3'
 #include 'lib/GWA2_Assembly.au3'
+#include 'lib/GWA2_Assembly_Chatlog.au3'
 #include 'lib/Utils.au3'
 #include 'lib/Utils-Agents.au3'
 #include 'lib/Utils-Storage.au3'
@@ -135,6 +136,8 @@ $run_options_cache['run.donate_faction_points'] = True
 $run_options_cache['run.buy_faction_scrolls'] = False
 $run_options_cache['run.buy_faction_resources'] = False
 $run_options_cache['run.collect_data'] = False
+$run_options_cache['run.go_offline'] = False
+$run_options_cache['run.flash_whisper'] = False
 $run_options_cache['team.automatic_team_setup'] = False
 ; Overrides on $run_options_cache for frequent usage
 Global $district_name = 'Random EU'
@@ -201,6 +204,8 @@ Func Main()
 		EndIf
 		; Authentication
 		Authentification($character_name)
+		If $run_options_cache['run.go_offline'] Then SetPlayerStatus(0)
+		If $run_options_cache['run.flash_whisper'] Then EnableWhisperFlash()
 		$runtime_status = 'RUNNING'
 	Else
 		MsgBox(0, 'Error', 'Unknown run mode: ' & $run_mode)
@@ -410,6 +415,7 @@ Func ReadConfigFromJson($jsonString)
 	$run_options_cache['run.sort_items'] = _JSON_Get($jsonObject, 'run.sort_items')
 	$run_options_cache['run.sort_items'] = _JSON_Get($jsonObject, 'run.sort_items')
 	$run_options_cache['run.collect_data'] = _JSON_Get($jsonObject, 'run.collect_data')
+	$run_options_cache['run.go_offline'] = _JSON_Get($jsonObject, 'run.go_offline')
 	$run_options_cache['run.donate_faction_points'] = _JSON_Get($jsonObject, 'run.donate_faction_points')
 	$run_options_cache['run.buy_faction_resources'] = _JSON_Get($jsonObject, 'run.buy_faction_resources')
 	$run_options_cache['run.buy_faction_scrolls'] = _JSON_Get($jsonObject, 'run.buy_faction_scrolls')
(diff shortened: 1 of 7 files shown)


BotsHub.au3
```diff
@@ -410,6 +415,7 @@ Func ReadConfigFromJson($jsonString)
 	$run_options_cache['run.sort_items'] = _JSON_Get($jsonObject, 'run.sort_items')
 	$run_options_cache['run.sort_items'] = _JSON_Get($jsonObject, 'run.sort_items')
 	$run_options_cache['run.collect_data'] = _JSON_Get($jsonObject, 'run.collect_data')
+	$run_options_cache['run.go_offline'] = _JSON_Get($jsonObject, 'run.go_offline')
```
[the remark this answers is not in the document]
buddyleex: Good catch copilot I need to add this to the config.

lib/GWA2_Assembly_Chatlog.au3
```diff
@@ -140,111 +130,179 @@ Func MemoryRevertDetour($fromLabel)
...
+		; separator is absent the event is not a real player whisper — skip it.
+		Local $msgPtr = GetChatMessagePtr()
+		If $msgPtr = 0 Then Return
+		Local $message = MemoryRead(GetProcessHandle(), $msgPtr, 'wchar[64]')
+		Local $separatorPos = StringInStr($message, 'Ĉ')
+		If $separatorPos = 0 Then Return
+		$sender = StringMid($message, 3, $separatorPos - 3)
+	EndIf
+
+	Call($chat_receive_function, $channel, $sender, '', 'No')
```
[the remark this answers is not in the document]
buddyleex: I'll add the limitation to the function commentary. No other functions leverage ChatLogPollCallback, but copilot is right that it would need to be expanded for further usage. However, the scaffolding is there, and  accurate. the copilot suggestions for text extraction and separators are incorrect.

lib/BotsHub-GUI.au3
```diff
@@ -635,6 +641,10 @@ Func GuiStartButtonHandler()
 		Case 'UNINITIALIZED'
 			$character_name = GUICtrlRead($gui_combo_characterchoice)
 			If (Authentification($character_name) <> $SUCCESS) Then Return
+			If $character_name <> '' Then
+				If $run_options_cache['run.go_offline'] Then SetPlayerStatus(0)
+				If $run_options_cache['run.flash_whisper'] Then EnableWhisperFlash()
+			EndIf
```
caustic-kronos: I suggest moving this to GeneralFarmSetup - that sounds like the kind of thing we'd want to do there.

lib/GWA2_Assembly_Chatlog.au3
```diff
@@ -1,44 +1,30 @@
 #include-once
 
+; Sentinel used by GWA2_Assembly.au3 to detect this module and call the Extend* hooks.
 Global Const $CHAT_LOG_STRUCT = 'dword;wchar[256]'
 
 Global $detours_map[]
 
 Global $chat_receive_function
```
caustic-kronos: I think this variable has a single value everywhere - can we remove the variable and use the function instead of using Call ?

lib/GWA2_Assembly_Chatlog.au3
```diff
@@ -1,44 +1,30 @@
...
-Global $chat_log_base_address
 Global $chat_log_counter_address
 Global $chat_message_channel_address
+Global $chat_message_ptr_address
+Global $chat_log_last_counter = 0
 
 Func ExtendAddPattern()
```
caustic-kronos: Let's rename this to be more explicit that we add a pattern for ChatLog.
Maybe AddChatLogScanPattern ?

lib/GWA2_Assembly_Chatlog.au3
```diff
@@ -140,111 +130,179 @@ Func MemoryRevertDetour($fromLabel)
...
 
+;------------------------------------------------------
+; Title...........:	_ChatLogOnExit
+; Description.....:	OnAutoItExitRegister handler. Reverts the ChatLog detour whenever
+;					BotsHub exits (normal close, error, or tray exit) so GW's code at
+;					ChatLogStart is restored to its original bytes. Without this, a
+;					subsequent BotsHub session would find a stale JMP at ChatLogStart and
+;					crash GW the next time a chat message arrives.
+;------------------------------------------------------
+Func _ChatLogOnExit()
```
caustic-kronos: For now there are no functions starting with _ in the Hub (with the exception of assembler).
Maybe we should have them - but for now for coherence sake let's remove the _ 

And then the question is: should we make all callback functions (onExit ? AdLib ? Parameter function like the fight functions ?) be written starting with a _ ?

BotsHub.au3
```diff
@@ -135,6 +136,8 @@ $run_options_cache['run.donate_faction_points'] = True
 $run_options_cache['run.buy_faction_scrolls'] = False
 $run_options_cache['run.buy_faction_resources'] = False
 $run_options_cache['run.collect_data'] = False
+$run_options_cache['run.go_offline'] = False
+$run_options_cache['run.flash_whisper'] = False
```
caustic-kronos, line 140: We need to add those options to the default configuration file

BotsHub.au3
```diff
@@ -201,6 +204,8 @@ Func Main()
 		EndIf
 		; Authentication
 		Authentification($character_name)
+		If $run_options_cache['run.go_offline'] Then SetPlayerStatus(0)
+		If $run_options_cache['run.flash_whisper'] Then EnableWhisperFlash()
```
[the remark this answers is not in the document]
caustic-kronos: Moving those calls to general farm setup will remove the need to have those function called in several places as well.

lib/GWA2_Assembly_Chatlog.au3
```diff
@@ -140,111 +130,179 @@ Func MemoryRevertDetour($fromLabel)
...
+	AdlibUnRegister('ChatLogPollCallback')
+	MemoryRevertDetour('ChatLogStart')
+EndFunc
+
 Func ChatLogSetEventCallback($chatReceive = '')
 	If $chatReceive <> '' Then
-		MemoryWriteDetourEx('ChatLogStart', 'ChatLogProc')
+		MemoryWriteDetourEx('ChatLogStart', 'ChatLogProc', 'ChatLogTrampoline')
+		$chat_log_last_counter = GetChatMessageCounter()
+		AdlibRegister('ChatLogPollCallback', 1000)
```
caustic-kronos: Just a warning that Adlib every second is quite often - it's a lot of interruptions.
In the case of a received message I'm not sure it's very important to have this kind of time resolution. I feel like 10s is largely enough (I'd even go lower personally) but maybe there are scenarios where it makes sense to have this kind of reactivity ?

caustic-kronos (changes requested):
Nice ! I can't wait to try those out. That will make a little bot to sell stuff very easy to do, only thing is to send trade message at repeated intervals and flash/make a sound on received whisper.
Added a few comments, that plus one or two valid Copilot comments and it's all good

BotsHub.au3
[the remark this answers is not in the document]
buddyleex: good call

lib/GWA2_Assembly_Chatlog.au3
```diff
@@ -140,111 +130,179 @@ Func MemoryRevertDetour($fromLabel)
...
-	Return 0
+	Local $sender = ''
+	If $channel = 'Received Whisper' Then
+		; Read the GW message string from the pointer stored by the hook.
+		; Zone transition messages also arrive on channel 14 as GW-encoded text tokens
+		; (e.g. 8102 0000) and will not contain the whisper separator 'Ĉ'. If the
+		; separator is absent the event is not a real player whisper — skip it.
+		Local $msgPtr = GetChatMessagePtr()
+		If $msgPtr = 0 Then Return
+		Local $message = MemoryRead(GetProcessHandle(), $msgPtr, 'wchar[64]')
```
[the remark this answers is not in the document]
buddyleex: This isn't correct. Bad Robot.

lib/BotsHub-GUI.au3
buddyleex: Yep I just noticed that I'm calling it twice, I'll move to General Farm Setup.

lib/GWA2_Assembly_Chatlog.au3
buddyleex: Sorry I can move it back. I'm just used to creating helper functions or call back functions as with the preceding underscore. I personally think they are helpful.

lib/GWA2_Assembly_Chatlog.au3
buddyleex: So I think this is where we are going to have major friction with this feature.

During testing, it was 100ms...I knew that you don't like such aggressive timers so I set it to 1000, but I never tested it assuming it would work. While fixing some of the issues with the PR just now, I noticed anything  greater than 100ms, just doesn't work. The pointer becomes stale or invalid by the time a second has passed. 

I'm going to set it back to 100ms with a comment so that all issues are resolved minus the poling of the ChatLog. Then  we can debug and see if this is purely a limitation with guild wars internal messaging buffer and/or if we can find a way to back the timers off without losing feature functionality.

lib/GWA2_Assembly_Chatlog.au3
buddyleex: Yeah that makes a lot more sense good call!

lib/GWA2_Assembly_Chatlog.au3
buddyleex: Ok new commits pushed.

I'll work on some debug tools to see if I can figure out what is going on and see if we can make the polling interval more conservative.

caustic-kronos: That's interesting.
Technically, I'm okay with a AdlibRegister at 100ms - if the bot doesn't do much next to that. So for a selling bot, that basically does nothing most of the time and just post a trade message every XX seconds, it doesn't matter much to have interruptions every 100ms.
For a bot running constantly doing things (especially with other AdlibRegister or busy wait style loops), such an AdlibRegister will cause issues though. I noticed it when trying to run Tasca with heal every 130ms or so - I started having weird unexplainable issues at every single run and the bot failed more despite having more reactive heal.

buddyleex: I figured out the issue reading from memory. This is a guild wars constraint, the buffer is cleared or updated every ~ 150ms. I've pushed a new commit. Instead of reading from the gw memory directly I've increased the buffer created by AssemblerCreateChatLog to read from instead and I've successfully tested up to 10s polling (currently set to 5s polling).

caustic-kronos: Nice fix. It leaves an issue though - we are using a simple buffer to write messages into, and a simple sentinel variable to get the channel it comes from.
Essentially if we receive a message from another channel after receiving the whisper, we'll overwrite it, won't we ?

We should probably add a little bit more of assembler:
```
_('WhisperCounter/4')
```
Then inside hook:
```
_('cmp dword[ebp+C],14')
_('jne skipWhisper')

_('mov eax,dword[WhisperCounter]')
_('inc eax')
_('mov dword[WhisperCounter],eax')

_('skipWhisper:')
```
Now, we count the whispers we receive. We only have to test whisperCounter != lastWhisperCounter to know whether or not we received a whisper.

caustic-kronos: That should be enough for our whisper flash needs.

If we wanted to improve this further we'd need to move message buffer to an event queue and make everything more complicated but we have no need for that here.

buddyleex: Good points on the limitations of the initial iteration. I implemented your fix and also included  WhisperData into the assembly. Simplified the PollCallback, and moved channel mapping to GetChannelName scaffolding for future implementations like alerting on all/trade chat like "WTB conset", etc.

caustic-kronos (approved):

PR #175, opened 2026-05-18
fix: correct KickAllHeroes parameter

buddyleex:


caustic-kronos (approved):
Thank you for this, let's merge !

PR #187, opened 2026-06-07
feat:	Update Dragon Moss Farm

Azgw:
The following changes were made:
- Add a R/P hero to the party. It casts some shouts to get better IMS, Winnowing and EoE. While we aggro and retreat to ball the enemies, it moves back towards Saint Anjeka's Shrine. The hero is optional and the farm still works without it.
- Change the casting order for spells before aggroing to reduce the time waiting for SF before killing.
- Change the pathing for doing aggro, which seems to reduce the change of missing the south-east group.

Overall it reduces the time of a run from ~68s to ~61s (testing done by running both versions for about an hour. I am not sure that "Stand Your Ground!" is super useful while "Can't Touch This!" is straight out useless. Maybe using a PR hero instead would be better (or just two heroes?)

I still have two issues:
- the bot will sometimes resign too fast and not pick all the loots (with all enemies being dead and not being in any immediate danger).
- sometimes after DC, we have a bad positionning and we leave one enemy alive. The issue is mitigated by the fact that the bot still loots and continues.

### DIFF
--- a/src/farms/DragonMoss.au3
+++ b/src/farms/DragonMoss.au3
@@ -60,7 +60,7 @@ Global Const $DM_RANGER_HERO_EDGE_OF_EXTINCTION = 7
 Global Const $DM_RANGER_HERO_WINNOWING 			= 8
 
 ; Order heros are added to the team
-Global Const $DM_RANGER_HERO		= 1
+Global $DM_RANGER_HERO	= 1
 
 Global $dm_farm_setup = False
 
@@ -81,7 +81,7 @@ Func SetupDragonMossFarm()
 	If TravelToOutpost($ID_SAINT_ANJEKAS_SHRINE, $district_name) == $FAIL Then Return $FAIL
 	SwitchMode($ID_HARD_MODE)
 	If SetupPlayerDragonMossFarm() == $FAIL Then Return $FAIL
-	If SetupTeamDragonMossFarm() == $FAIL Then Return $FAIL
+	If SetupTeamDragonMossFarm() == $FAIL Then Warn('Could not add ranger hero to team. Continuing without hero')
 	GoToDrazachThicket()
 	MoveTo(-11100, 19700)
 	Move(-11300, 19900)
@@ -118,7 +118,6 @@ Func SetupTeamDragonMossFarm()
 		SetHeroBehaviour(1, $ID_HERO_AVOIDING)
 		RandomSleep(150)
 	Else
-		Warn('Could not add ranger hero to team. Continuing without hero')
 		$DM_RANGER_HERO = 0
 		Return $FAIL
 	EndIf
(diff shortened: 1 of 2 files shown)


src/farms/DragonMoss.au3
```diff
@@ -111,37 +142,70 @@ EndFunc
 Func DragonMossFarmLoop()
 	If GetMapID() <> $ID_DRAZACH_THICKET Then Return $FAIL
 
+	; Speed boosting and moving to the shrine
+	If $DM_RANGER_HERO > 0 Then
+		UseHeroSkill($DM_RANGER_HERO,$DM_RANGER_HERO_INCOMING)
+		RandomSleep(50)
```
caustic-kronos: No need to add a sleep here - Incoming has no cast time, and we don't make hero cast another skill right after so no need to make sure the first one was casted before doing the second.

src/farms/DragonMoss.au3
```diff
@@ -93,6 +105,25 @@ Func SetupPlayerDragonMossFarm()
...
+		RandomSleep(150)
+		LoadSkillTemplate($DM_RANGER_HERO_SKILLBAR, 1)
+		RandomSleep(150)
+		DisableAllHeroSkills(1)
+		SetHeroBehaviour(1, $ID_HERO_AVOIDING)
+		RandomSleep(150)
+	Else
+		Warn('Could not add ranger hero to team. Continuing without hero')
+		$DM_RANGER_HERO = 0
+		Return $FAIL
```
caustic-kronos, line 122: Returning $FAIL here means that we get fail at line 84 - which makes us return $FAIL overall and stop farm - so currently farm should not work without hero.
We can keep the warning but return $SUCCESS - that way we can really run without hero too.

caustic-kronos (changes requested):
Hello !
Thanks for making those changes, good stuff overall.
Two tiny issues but nothing too bad.
Will merge after fixes

src/farms/DragonMoss.au3
Azgw: Would it make sense to have the function still return $FAIL but having the call at line 84 not return $FAIL?

caustic-kronos: Yes it makes complete sense, you can go with that solution.

If you make this function (the called function) return fail, you communicate "We could not set up the team as we wanted".
Then it's up to the caller function to decide what to do with this result. In our case it would be "Sure, team setup failed, but this bot doesn't need the team setup to have worked anyway, so we continue."

Azgw:
Hello again,
I've addressed your comment and the `SetupDragonMossFarm` function on a character without access to a ranger hero.

caustic-kronos (approved):
Great, let's merge 😄

PR #208, opened 2026-07-29
fix: honor alcohol storage setting

Wellwisher533:
## Summary
- route alcohol through the dedicated `Store items.Alcohols` setting
- preserve the existing full-stack-only storage behavior
- leave other consumables controlled by `Store items.Consumables`

## Problem
`IsConsumable()` includes alcohol, so the existing storage branch always reads `Store items.Consumables` for alcohol and never reads the dedicated alcohol option.

## Validation
- `git diff --check` passes
- AutoIt control flow matches the existing alcohol-specific pickup handling
- full-project `Au3Check` remains blocked by pre-existing errors in unrelated upstream code


Wellwisher533:
Closing pending successful end-to-end compilation and runtime validation.

PR #214, opened 2026-08-06
Add CoF Sin CA: Assassin Shadow Form farmer for Cathedral of Flames

phiwi:
Alternative to the Dervish VoS farmer — for players without a Dervish.

### Performance (190 runs overnight)
- **~80% success rate** (timeouts count as success — loot still collected)
- **Typical run: 1:45–2:50** (comparable to Dervish)
- **Timeout runs (10 min):** still pick up all loot before returning
- **Fails:** mostly Ghoul Distracting Blow interrupts (RNG), not build issues
- **Overnight loot:** 2 Diessa Chalices + 3 Golden Rin Relics

### Build: `Owpk8xjYaqWEPiOzB0YozV3XNImI`

| Slot | Skill | Purpose |
|------|-------|---------|
| 1 | Deadly Paradox | 33% faster Assassin skill recharge |
| 2 | Shadow Form | Spell immunity + 5 DR/enchant |
| 3 | Shroud of Distress | 75% block + 7 regen under 50% HP |
| 4 | Crippling Victory | Cripple + AoE earth damage |
| 5 | Reap Impurities | Condition removal → AoE holy damage |
| 6 | Grenth's Aura | Life steal on scythe hits |
| 7 | Critical Agility | Perma-IAS (self-renews on crit) |
| 8 | Signet of Mystic Speed | Instant self-enchant casting |

### Optimizations
- **DP→SF casting:** Deadly Paradox 500ms *before* Shadow Form — SF snapshots 33% faster recharge (20s instead of 30s)
- **AdlibRegister perma loop:** independent 200ms SF/DP maintenance prevents gaps
- **10-min combat timeout** with guaranteed loot pickup before returning
- **Pre-combat perma check:** SF/DP priority before every blocking skill (CV/Reap)

### DIFF
--- a/BotsHub.au3
+++ b/BotsHub.au3
@@ -51,6 +51,7 @@ Opt('MustDeclareVars', True)
 #include 'lib/Build_PW_Heroic-Refrain.au3'
 
 #include 'src/farms/CoF.au3'
+#include 'src/farms/CoF Sin CA.au3'
 #include 'src/farms/Corsairs.au3'
 #include 'src/farms/DragonMoss.au3'
 #include 'src/farms/DrakeFlesh.au3'
@@ -539,6 +540,7 @@ Func FillFarmMap()
 	AddFarmToFarmMap(	'Asuran',						AsuranTitleFarm,				5,					$ASURAN_FARM_DURATION)
 	AddFarmToFarmMap(	'Boreal',						BorealChestFarm,				5,					$BOREAL_FARM_DURATION)
 	AddFarmToFarmMap(	'CoF',							CoFFarm,						5,					$COF_FARM_DURATION)
+	AddFarmToFarmMap(	'CoF Sin CA',					CoFSinCAFarm,					5,					$COFSINCA_FARM_DURATION)
 	AddFarmToFarmMap(	'Corsairs',						CorsairsFarm,					5,					$CORSAIRS_FARM_DURATION)
 	AddFarmToFarmMap(	'Deldrimor',					DeldrimorFarm,					10,					$DELDRIMOR_FARM_DURATION)
 	AddFarmToFarmMap(	'Drake Flesh',					DrakeFleshFarm,					5,					$DRAKE_FLESH_FARM_DURATION)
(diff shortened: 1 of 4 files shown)


caustic-kronos:
Hello, thanks for this PR ! 🥳 
I have to look deeper into it though, there are several things that I think need to be done.
1- Merge it with existing CoF: if we start making a file per build, it will soon become unmanageable, and from what I could see, there is a solid base between both, so we probably can merge them and refactor some code.
2- You are using HeroHasTemplate, which would be an amazing addition, not only to this farm, but to all other farms present in the Hub. Unfortunately it's not present in the Hub, it must be one of your functions. Add the function to the PR or remove it from the bot.
3- It's good to have file logging, but we need to make it optional, disabled by default, otherwise heavy users of the bot will be growing a giant file in their folder.
4- AdlibRegister should be avoided if possible. It has significant performance impact on the bot, especially at high frequencies, or if several functions are registered.

phiwi:
Thanks for the review! All four points addressed:

1. **Merged** — `CoF.au3` now handles Dervish + Assassin in one file via profession dispatch. Single dropdown entry.
2. **`HeroHasTemplate`** — added to `lib/Utils.au3`, available for all farms (removed duplicate from SupportTeam.au3).
3. **CSV logging** — gated behind `$COF_SIN_DEBUG_LOG = False`, disabled by default.
4. **`AdlibRegister` removed** — `MaintainCoFSinPerma()` is called inline 2× per combat iteration; the 200ms Adlib was redundant.

The PR branch (`cof-sin-farms`) has been updated with all fixes. Ready for another look!

caustic-kronos:
I pushed CoF A in 268e6db82e7cd47199a1474e1933ece3a70e82cc
Ended up adapting quite a bit, and removed all logging, it makes code too noisy, those logs belong more into a debug branch than on master.
Corrected HeroHasTemplate it wasn't working well, now it should do the trick, thank you 😄 

CoF A is still not very efficient, I find. Try it and tell me if I fucked something up or if it's normal. I guess the reduction of damage from SF + the 10s of attacks being on CD doesn't make the farm very efficient.

caustic-kronos:
Since A is now usable in CoF, this PR can be closed.
Thank you so much, cheers 😁

PR #215, opened 2026-08-06
Pull and builds fixed

kneemant:
fixed pull, added all builds for all classes

### DIFF
--- a/src/missions/UnderworldPantheon.au3
+++ b/src/missions/UnderworldPantheon.au3
@@ -118,8 +118,6 @@ EndFunc
 
 ;~ Single-pass: Spawn spirits, wair for exactly 2 aatxe, pull, kill, loot
 Func ClearTheChamberUnderworldPantheon()
-	$uw_pantheon_run_timer = TimerInit()
-
 	Info('Place Spirits for block')
 	MoveTo(1376, 7418)
 	UseSkillEx($UW_FARM_BLOODSONG)
--- a/src/missions/UnderworldPantheon.au3
+++ b/src/missions/UnderworldPantheon.au3
@@ -184,10 +184,7 @@ Func WaitForExactlyTwoAatxe()
 		If IsPlayerDead() Then Return $FAIL
 	WEnd
 
-	If $foesCount <= 1 Then
-		Info('Only ' & $foesCount & ' Aatxe in range - targeting nearest enemy and pulling anyway')
-		GetNearestEnemyToAgent(GetMyAgent(), 2000)
-	EndIf
+	If $foesCount <= 1 Then Info('Only ' & $foesCount & ' Aatxe in range - targeting nearest enemy and pulling anyway')
 
 	Return $SUCCESS
 EndFunc
--- a/src/missions/UnderworldPantheon.au3
+++ b/src/missions/UnderworldPantheon.au3
@@ -27,16 +27,16 @@
 
 
 ; ==== Constants ====
-Global Const $UW_PANTHEON_A_SKILLBAR					= 'OwhjAyi84QlTfT+gXTRbnNVTQT'
-Global Const $UW_PANTHEON_E_SKILLBAR					= 'OghjwMgsITlTfT+gXTRbnNVTQT'
-Global Const $UW_PANTHEON_D_SKILLBAR					= 'OgijAyiM7QlTfT+gXTRbnNVTQT'
-Global Const $UW_PANTHEON_ME_SKILLBAR					= 'OQhjAyiMwQlTfT+gXTRbnNVTQT'
-Global Const $UW_PANTHEON_MO_SKILLBAR					= 'OwgjAyiM0QlTfT+gXTRbnNVTQT'
-Global Const $UW_PANTHEON_N_SKILLBAR					= 'OAhjAyisxQlTfT+gXTRbnNVTQT'
-Global Const $UW_PANTHEON_P_SKILLBAR					= 'OQijAyiM6QlTfT+gXTRbnNVTQT'
-Global Const $UW_PANTHEON_RA_SKILLBAR					= 'OggjAyi81QlTfT+gXTRbnNVTQT'
-Global Const $UW_PANTHEON_RT_SKILLBAR					= 'OACjAyiM5QlTfT+gXTRbnNVTQT'
-Global Const $UW_PANTHEON_W_SKILLBAR					= 'OQgjAyic0QlTfT+gXTRbnNVTQT'
+Global Const $UW_PANTHEON_A_SKILLBAR					= 'OwhiIygMVO9NdN5De2EtVNBNB'
+Global Const $UW_PANTHEON_E_SKILLBAR					= 'OghiIygMVO9NdN5De2EtVNBNB'
+Global Const $UW_PANTHEON_D_SKILLBAR					= 'OgiiIygMVO9NdN5De2EtVNBNB'
+Global Const $UW_PANTHEON_ME_SKILLBAR					= 'OQhiIygMVO9NdN5De2EtVNBNB'
+Global Const $UW_PANTHEON_MO_SKILLBAR					= 'OwgiIygMVO9NdN5De2EtVNBNB'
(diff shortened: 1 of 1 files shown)


src/missions/UnderworldPantheon.au3
```diff
@@ -27,16 +27,16 @@
...
+Global Const $UW_PANTHEON_A_SKILLBAR					= 'OwhiIygMVO9NdN5De2EtVNBNB'
+Global Const $UW_PANTHEON_E_SKILLBAR					= 'OghiIygMVO9NdN5De2EtVNBNB'
+Global Const $UW_PANTHEON_D_SKILLBAR					= 'OgiiIygMVO9NdN5De2EtVNBNB'
+Global Const $UW_PANTHEON_ME_SKILLBAR					= 'OQhiIygMVO9NdN5De2EtVNBNB'
+Global Const $UW_PANTHEON_MO_SKILLBAR					= 'OwgiIygMVO9NdN5De2EtVNBNB'
+Global Const $UW_PANTHEON_N_SKILLBAR					= 'OAhiIygMVO9NdN5De2EtVNBNB'
+Global Const $UW_PANTHEON_P_SKILLBAR					= 'OQiiIygMVO9NdN5De2EtVNBNB'
+Global Const $UW_PANTHEON_RA_SKILLBAR					= 'OggiIygMVO9NdN5De2EtVNBNB'
+Global Const $UW_PANTHEON_RT_SKILLBAR					= 'OACiIygMVO9NdN5De2EtVNBNB'
+Global Const $UW_PANTHEON_W_SKILLBAR					= 'OQgiIygMVO9NdN5De2EtVNBNB'
```
caustic-kronos, line 39: Those new builds are not correct - I changed order of skills on purpose so that they are in the order we cast them: we need to cast spirits from longest lasting to shortest lasting to garantee that they stay up as long as possible during the fight. Which is why Vampirism is 3rd (100+s), Signet of Spirits is 4th (60s), Anguish 5th (40s) and Blind 6th (30s).
This change should be reverted

src/missions/UnderworldPantheon.au3
```diff
@@ -63,6 +63,8 @@ Global Const $UW_AATXE_CHECK_INTERVAL = 1000
 Global Const $UW_AATXE_SKILL_RECAST_INTERVAL = 5000
 
 Global $uw_pantheon_farm_setup = False
+; Timer for the 2-minute run timeout, started at the beginning of each run
+Global $uw_pantheon_run_timer = 0
```
caustic-kronos:
```suggestion
```
We don't need this timer because we can use the global timer that is used in all bots.

src/missions/UnderworldPantheon.au3
```diff
@@ -116,8 +118,16 @@ Func SetupPlayerUnderworldPantheonFarm()
 EndFunc
 
 
+;~ Returns True once the run has exceeded the 2-minute limit
+Func IsPantheonRunTimedOut()
+	Return TimerDiff($uw_pantheon_run_timer) >= $MAX_UW_FARM_PANTHEON_DURATION
+EndFunc
+
+
```
caustic-kronos:
```suggestion
```
Since we already have a way to check time, we can also remove this function.

src/missions/UnderworldPantheon.au3
```diff
@@ -116,8 +118,16 @@ Func SetupPlayerUnderworldPantheonFarm()
...
 ;~ Single-pass: Spawn spirits, wair for exactly 2 aatxe, pull, kill, loot
 Func ClearTheChamberUnderworldPantheon()
+	$uw_pantheon_run_timer = TimerInit()
+
```
caustic-kronos:
```suggestion
```

src/missions/UnderworldPantheon.au3
```diff
@@ -128,6 +138,7 @@ Func ClearTheChamberUnderworldPantheon()
 	Sleep(500)
 	; First 3 spirits have very long duration, so we can wait for energy to be maxxed again
 	While DllStructGetData(GetMyAgent(), 'EnergyPercent') < 0.99
+		If IsPantheonRunTimedOut() Then Return $FAIL
```
caustic-kronos:
```suggestion
		If CheckStuck('UW Pantheon - Waiting for max energy', $MAX_UW_FARM_PANTHEON_DURATION) == $FAIL Then Return $FAIL
```

src/missions/UnderworldPantheon.au3
```diff
@@ -141,9 +152,10 @@ Func ClearTheChamberUnderworldPantheon()
 	If WaitForExactlyTwoAatxe() == $FAIL Then Return $FAIL
 	; Wait for enough energy to cast painful bond
 	While GetEnergy() < 15
+		If IsPantheonRunTimedOut() Then Return $FAIL
```
caustic-kronos:
```suggestion
		If CheckStuck('UW Pantheon - Waiting for energy', $MAX_UW_FARM_PANTHEON_DURATION) == $FAIL Then Return $FAIL
```

src/missions/UnderworldPantheon.au3
```diff
@@ -165,17 +177,26 @@ Func ClearTheChamberUnderworldPantheon()
...
+;~ - exactly 2: proceed normally
+;~ - 0 or 1: Aatxe are too far away to be counted in range - target the nearest enemy and pull anyway
 Func WaitForExactlyTwoAatxe()
 	Local $foesCount = CountFoesInRangeOfAgent(GetMyAgent(), 2000)
-	While $foesCount <> 2
-		Info('3 Aatxes nearby - wait until 1 leaves')
+	While $foesCount > 2
+		Info('More than 2 Aatxes nearby - wait until 1 leaves')
 
+		If IsPantheonRunTimedOut() Then Return $FAIL
```
caustic-kronos:
```suggestion
		If CheckStuck('UW Pantheon - Waiting for 2 Aaxtes or less', $MAX_UW_FARM_PANTHEON_DURATION) == $FAIL Then Return $FAIL
```

src/missions/UnderworldPantheon.au3
```diff
@@ -165,17 +177,26 @@ Func ClearTheChamberUnderworldPantheon()
...
 		Sleep($UW_AATXE_CHECK_INTERVAL)
 		$foesCount = CountFoesInRangeOfAgent(GetMyAgent(), 2000)
 		If IsPlayerDead() Then Return $FAIL
 	WEnd
+
+	If $foesCount <= 1 Then
+		Info('Only ' & $foesCount & ' Aatxe in range - targeting nearest enemy and pulling anyway')
+		GetNearestEnemyToAgent(GetMyAgent(), 2000)
+	EndIf
```
caustic-kronos:
```suggestion
	If $foesCount <= 1 Then Info('Only ' & $foesCount & ' Aatxe in range - targeting nearest enemy and pulling anyway')
```
The GetNearestEnemyToAgent call does nothing here (we don't use its result)

src/missions/UnderworldPantheon.au3
```diff
@@ -185,6 +206,8 @@ Func WaitUntilAatxeDead()
 	Local $lastSkillCast = TimerInit()
 	Local $foesCount = CountFoesInRangeOfAgent(GetMyAgent(), $RANGE_SPELLCAST)
 	While $foesCount > 0
+		If IsPantheonRunTimedOut() Then Return $FAIL
```
caustic-kronos:
```suggestion
		If CheckStuck('UW Pantheon - Waiting for Aaxtes to be dead', $MAX_UW_FARM_PANTHEON_DURATION) == $FAIL Then Return $FAIL
```

caustic-kronos (changes requested):
Hey, good idea to add timeout checks to prevent bot from stalling !
But there is already a mechanism present in the Hub to do this check without every bot needing to define a timer and a specific function to make the check.
So I replaced your calls to it by the existing system.

Additionally, I believe your changes on skillbars are incorrect, unless I missed something. So that part should be reverted.

caustic-kronos:
I moved your changes along with the corrections to this PR:
https://github.com/caustic-kronos/BotsHub/pull/217
Closing this one. Thank you ❤️

Issue #211, opened 2026-08-01
Vaettir Farms observations + improvements


Wellwisher533:
Hi, love the bot. Been using it a lot during the previous event for the first time, mostly farming Vaettirs. I found a few things and applied some of them locally, but wanted to get your take before submitting some of them as PRs.

**A/Me**
* Most safe, reliable farm, but can die on the route through Bjora Marches. I found a safer route that takes slightly longer but have had 0 deaths. 
* The A/Me can also die during pickup. It seems to treat (no Vaettirs within aggro range) as success but every once in a while, it fails the final balling somewhat and Vaettirs don't show up closely, but _they do follow._ It's noticeable when it declares "Starting to kill" to "Vaettirs all cleared." What happens is that the A/Me tries to pick up loot (when there's none), realize it's "done" and tries to rezone, only for Vaettirs to chase them down and kill. I added that whenever zero Vaettirs reached the killing spot, it abandons the run immediately and travels back to Longeye's Ledge to avoid death, but it can be as easily as tweaking it to "rezone, but assume danger so be ready to fight on your way to rezone."

**E/Me**
* The build has a hard time running through Bjora Marches. I don't really know how to fix it outside of the safer route, but even then, wurms can show up and ruin your day. For extreme safety (if you're going for Survivors), I think the only way is to log out as soon as you detect a wurm (i.e., get knocked down).
* The build gets stuck a lot. An experienced human player can ball up every Vaettirs and not get body-blocked, but it's hard for a bot. The PvX wiki suggests doing two separate balls -> kill events, but the current aggro/ball path is the same as A/Me. When the bot gets stuck, it waits for 100s before trying to kill. That's a really long wait. I tweaked it to 15s, and even then it was taking forever to complete vs. A/Me because it would get body-blocked/stuck at least 2x a run. I am not sure how to fix this, even though E/Me is supposed to be the fastest of non-Dervish builds.
* The E/Me assumes the player must have the Kurzick version of the Elemental Lord but never checks if the player has the Luxon version but not the Kurzick version, or that the Luxon version might be a higher rank.

**Mo/A**
[...]

PR #218, opened 2026-08-10
fix: stabilize 55 HP Monk Vaettir protection

Wellwisher533:
## Problem

The 55 HP Monk could complete individual Vaettir runs, but it was not reliable over an extended farming session. In one unattended test before these changes, it completed 19 runs but then entered 65 failed/death cycles. Repeated deaths and recovery made the effective farming pace drastically worse.

The failures came from several interacting issues:

- Protection refreshes were gated by nearby-enemy checks, allowing Shadow Form and Protective Spirit to expire during brief range gaps in both Jaga Moraine and the Bjora Marches run.
- Deadly Paradox, Protective Spirit, and Shadow Form were submitted inside a narrow recharge window. Closely spaced commands could be ignored or replace a queued cast.
- A Norn Hunting Party blessing can randomly grant Strength of the Norn (+100 maximum Health), breaking the 55 HP damage cap.
- After a death, the Monk could continue with death penalty, which makes the 55 HP setup effectively unusable and leads to repeated failures.

## Changes

- Give the Monk a dedicated Jaga protection transaction:
  - arm Deadly Paradox earlier;
  - submit Protective Spirit in a ping-adjusted window;
  - wait for Shadow Form recharge plus a small ping-adjusted settle period;
  - submit each skill only once and confirm client acceptance from its recharge state.
- Maintain Monk protection continuously in Jaga Moraine instead of tying it to enemy range.
- Keep Bjora Marches on a separate protection routine and remove the enemy-range gate for the Monk only. The route and other professions remain unchanged.
- Skip Norn Hunting Party on the 55 HP Monk so the random maximum-Health blessing cannot invalidate the build.
- Return the Monk to town after any death to clear death penalty before another attempt.

## Result

The purpose of this change is reliable farming throughput, not title acquisition. After the final fix, the same Monk completed 92 consecutive farm runs without another death, including three storage cycles and five successful Bjora crossings. The character also earned and exceeded Legendary Survivor during the run; that is useful evidence that the farm remained consistent over a long session, rather than the goal of the bot.

One earlier backgrounded run did fail before that 92-run streak, so this does not claim mathematical impossibility of failure. It changes the behavior from repeated failure loops to sustained unattended farming in the observed testing.

## Validation

- Replayed 991 recorded protection timing traces: 991 safe, 0 unsafe.
- Synthetic normal/background timing scenarios pass.
- Source-contract checks pass for cast ordering, one-shot submission, map isolation, continuous protection, Norn-blessing exclusion, and profession isolation.
- The AutoIt checker reports no findings in `src/farms/Vaettirs.au3`; existing shared-library checker errors remain outside this change.

Related to #211.

### DIFF
--- a/src/farms/Vaettirs.au3
+++ b/src/farms/Vaettirs.au3
@@ -264,7 +264,14 @@ Func RunAcrossBjoraMarches($X, $Y)
 		If IsPlayerDead() Then Return $FAIL
 		$target = GetNearestEnemyToAgent($me)
 
-		If GetDistance($me, $target) < 1300 And GetEnergy() > 20 Then VaettirsCheckBuffs()
+		; The 55 HP Monk must stay protected throughout Bjora. A distant enemy can
+		; remain dangerous after leaving the original 1300-unit detection gate.
+		; Preserve the original gate for every other profession.
+		If $vaettirs_player_profession == $ID_MONK Then
+			VaettirsCheckBuffs()
+		ElseIf GetDistance($me, $target) < 1300 And GetEnergy() > 20 Then
+			VaettirsCheckBuffs()
+		EndIf
 
 		If $vaettirs_player_profession <> $ID_ELEMENTALIST Then
 			$me = GetMyAgent()
@@ -292,9 +299,12 @@ Func VaettirsFarmLoop()
 
 	If IsPlayerDead() Then
 		$result = $FAIL
-		; Max malus - we return to the outpost and rerun
-		If IsPlayerAtMaxMalus() Then
-			Warn('Reached max death malus, restarting the farm setup')
+		; Any death penalty makes the 55 HP Monk effectively a 1 HP build. Return
+		; immediately to town to clear it; other professions retain max-DP recovery.
+		If $vaettirs_player_profession == $ID_MONK Or IsPlayerAtMaxMalus() Then
+			Warn($vaettirs_player_profession == $ID_MONK ? _
+				'Monk died; returning to town to clear death penalty' : _
+				'Reached max death malus, restarting the farm setup')
 			TravelToOutpost($ID_LONGEYES_LEDGE, $district_name)
 			$vaettirs_farm_setup = False
 			Return $FAIL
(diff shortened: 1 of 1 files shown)


caustic-kronos:
Hello, thank you for taking the time to work on this PR and for the detailed testing and validation 🙂

After looking through the changes, I will not merge this PR. This isn't because the work is poorly done; the additional reliability simply isn't worth the trade-offs for this farm.

There are a few reasons for that:

- Monk already has a very high success rate in the Vaettirs farm (around 95%), and is able to complete both Norn and Survivor titles through the farm. The improvements from this PR would only add a few percentage points of reliability, at the cost of longer farm times, losing Norn title points, and increased code complexity and maintenance. The main user benefiting from these changes would be someone running a Dhuum character, rather than the typical Monk user, and supporting Dhuum characters is not a goal for this farm.
- The improvements come with a fairly substantial amount of additional code: several new functions, additional branches in existing functions, and more specialized timing logic. The farm logic should be kept as generic and straightforward as possible. In this case, the added specialization makes the code harder to maintain and understand without enough corresponding benefit 😅
- Monk can already continue farming with death penalty, even when effectively reduced to 1 HP. Since that works in practice, returning to the outpost after a death would reduce farming efficiency rather than improve anything.
- Strength of the Norn is a relatively uncommon outcome, and even when it occurs it doesn't necessarily result in a failed run. Removing the blessing entirely means giving up the opportunity to gain Norn points on every run in order to avoid a relatively small risk. To me, that is a net loss for the existing use case.

So while I agree that the changes can make Monk runs more consistent, I don't think that consistency is valuable enough here to justify the additional complexity and the loss of existing functionality.

For those reasons, I'm going to leave this PR unmerged. I know that's disappointing after putting work into it, and I do appreciate the effort and the thorough validation you put into the change. I hope you don't take the rejection as a criticism of the work itself; it's primarily a decision about the project's scope and the trade-offs I'm willing to make in this particular farm.

Thank you again for taking the time to investigate this and submit the PR.

Wellwisher533:
Minus the Norn titles and such changes, I was not able to actually farm with the Monk due to the tight requirement needed to get Shadow Form up. It was actually failing at around 70-80% of the time for me after leaving it all night. I even made sure I got the latest branch. Maybe something my setting didn't match up, but would recommend someone else run the bot on their side to confirm this is a me only problem vs. a general problem. Otherwise, the monk farm might not work at all.

I even made sure to be in the same district / area as my local area, so it's not a ping issue. I asked it to monitor over night and the worst pings get to 200 ms, which is substantial, but the failure didn't come from high pings, but in how the bot schedules the skill casts. It simply doesn't leave enough time to cast Shadow Form, and it does not ensure that it's actually casted.

And of course, no worries about the rejection! It makes sense for the project in general to reduce complexity where it's not needed. But my experience with the base farm is that it was failing way more than it was succeeding to the point where I couldn't use it reliably, and thus had to build this. If that's something local to my branch, then that's that. I am under the assumption that it's impacting everyone in general.

PR #226, opened 2026-08-27
Support Infinite Identification Kit

Wellwisher533:
## Summary
- register the Infinite Identification Kit (model ID `38620`)
- prefer it in the shared identification-kit lookup used by inventory management
- treat it as satisfying mid-run kit requirements and skip purchasing disposable identification kits while it is present
- include it in the general-item catalog

This applies to farms using BotsHub's shared `IdentifyItem` / `IdentifyItems` inventory flow; no farm-specific code is changed.

## Verification
- focused static behavior checks cover the model ID, shared lookup priority, mid-run recognition, purchase suppression, and permanent-kit return path
- `git diff --check` passes
- the repository-wide AutoIt checker reports the same existing 44 errors and 72 warnings on the untouched upstream commit and this branch; this change introduces no additional diagnostics

Infinite Superior Salvage Kit support is intentionally not included in this PR and will be discussed separately.


Wellwisher533:
Follow-up / coordination note for the other new utility item:

We also wanted to add the **Infinite Superior Salvage Kit** (model ID `38621`), but deliberately left it out of this PR because its flow is not yet reliable.

The observed in-game materials flow for an armor salvageable with an upgrade is:

1. use the kit on the item;
2. a salvage chooser lists the rune/insignia and **Crafting Materials**;
3. select **Crafting Materials**;
4. click **Salvage**;
5. a separate **Yes / No** destructive confirmation appears;
6. click **Yes**, then verify the source item is gone/materials appeared.

This differs from the existing basic salvage-kit path. We tried a narrowly scoped BotsHub prototype and then a separate read-only UI diagnostic. The prototype was not reliable, and the Guild Wars client crashed during testing. In the diagnostic, a baseline snapshot before choosing an item succeeded, but querying the UI frame tree while the new salvage chooser was visible crashed `Gw.exe` before the chooser snapshot could be saved. We stopped rather than continue guessing against live accounts.

If you are already implementing support for model `38621`, please let us know and we will stop to avoid duplicating work. If not, any pointers on the supported/stable AutoIt frame or UI-message path for this chooser and its nested confirmation would be appreciated; we can continue testing once there is a safer direction.

caustic-kronos:
Hello,

Supporting those new items is definitely on my todo-list. I've not been home for 2 weeks so couldn't do a thing, but yes, infinite ID kit will be an easy addition once I come back.

Infinite salvage is more complicated, it has the same issue "advanced" salvage kits have. I stopped working on that more than a 1 year ago, so you are very welcome to  try to make it work.
Unfortunately I have no pointer, I had the same issue as you - game crashing. There might be other bots which managed to enable mod salvage, if so, taking example on them could work.

Wellwisher533:
I already got infinite ID kit to work (this PR) so let me know if this looks good. Otherwise, feel free to take what is written here and repurpose it.

Okay, let me ask around. I tested infinite salvage kits for an hour+ and couldn't get it to work.

caustic-kronos (approved):
I'm back home, can finally start merging things.
This looks good already, let's merge

PR #227, opened 2026-08-27
Add shield attribute filters to loot profiles

Wellwisher533:
## Summary

This adds attribute filtering for shields in loot profiles.

Shield settings are now grouped as:

- Strength
- Tactics
- Leadership
- Command
- Motivation
- Other

Each group keeps the existing Req 0 through Req 13 choices. The same lookup is used for pickup, salvage, sell, and store decisions. Non-shield weapon behavior is unchanged.

Existing requirement-only profiles are expanded in memory when loaded, so their current choices initially apply to every shield attribute. If a profile contains both formats, explicit attribute-specific choices take precedence. Missing legacy requirement values default to unchecked. Saving the profile writes the new nested structure.

`Other` is the fallback for any shield attribute outside the five normal PvE drop attributes. Existing valuable-item pickup overrides and the existing green-item rules are unchanged.

## Testing

- Ran an AutoIt runtime harness covering all five named attributes, the `Other` fallback, legacy migration, partial legacy settings, and mixed-format precedence.
- Built the loot tree from the shipped default profile in a hidden AutoIt GUI, serialized it, reloaded it, and verified all 1,512 applicable shield attribute/requirement leaves across pickup, salvage, sell, and store.
- Confirmed all four actions use the shared resolver and non-shield weapons retain the old requirement-only path.
- Compared the complete `BotsHub.au3` AutoIt checker output with the same upstream commit. Both report the same existing 44 errors and 72 warnings, with no diagnostics in the new code.
- `git diff --check` passes.

### DIFF
--- a/BotsHub.au3
+++ b/BotsHub.au3
@@ -388,9 +388,56 @@ Func LoadLootConfiguration($filePath)
 	$loot_configuration = StringTrimRight($loot_configuration, 5)
 	Local $jsonLootOptions = LoadLootOptions($filePath)
 	FillInventoryCacheFromJSON($jsonLootOptions, '')
+	ExpandShieldRequirementSettings()
 	BuildInventoryDerivedFlags()
 	RefreshValuableListsFromCache()
 	Info('Loaded loot configuration at ' & $filePath)
+EndFunc
+
+
+;~ Expand legacy Shield.Req N settings into attribute-specific shield settings.
+;~ The loot tree is generated from this cache, so existing profiles gain the new
+;~ hierarchy without requiring a destructive file migration.
+Func ExpandShieldRequirementSettings()
+	Local Static $configurationRoots[] = ['Pick up items', 'Salvage items', 'Sell items', 'Store items']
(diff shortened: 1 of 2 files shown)


caustic-kronos:
Hello, thank you for the PR !

I think we can definitely add this functionality for shields. I’d also like to eventually have the same option for staves, scepters, and foci, but I’m afraid that supporting all of them would make the decision tree too complex. So, keeping it limited to shields for now makes sense to me.

That said, I think we should modify the configuration files as part of this change. I understand the intent behind this PR: avoiding making existing configuration files obsolete is a thoughtful approach. However, I think the resulting compatibility code would make the implementation less maintainable in the long run. Ultimately, I’d rather accept the breaking change and update the configuration files accordingly. Sorry to users who would be affected by that.

There’s also another PR that was just opened which reverses the whole decision tree. I need to take a look at that one before merging anything, since it may affect how we want to approach this change. So, unfortunately, I’ll have to leave this PR paused for now. I’ll come back to it once I’ve reviewed the other PR.

PR #230, opened 2026-09-04
feat: Salvage into components option

krumpyzoid:
## Summary

New inventory option **Salvage into components** (Options tab, `run.salvage_into_components`, default off). When enabled, the valuable mods, inscriptions, runes and insignias are salvaged out of items that are not worth keeping for themselves, the components are stored like other valuables, and the stripped item then goes through the usual salvage/sell flow.

- Component detection reuses the *Keep components* selections of the loot configuration, split by salvage slot (prefix/insignia 0, suffix/rune 1, inscription 2): `GetValuableComponentIndices`, `ContainsAnyStruct`, `CreateValuableArmorUpgradesArray`, `CreateValuableModsBySlotByWeaponTypeMap`.
- `ShouldSalvageComponents` only strips items failing the keep rules; `ShouldKeepWeapon` gains a `$checkUpgrades` switch so weapons kept for their skin, requirement or stats stay untouched.
- Needs an expert or superior salvage kit: `GetExpertSalvageKit` buys an expert kit in town when needed and allowed, `GetSalvageKitUses` reads the remaining uses. The kit choice window is answered directly by packets in `SalvageComponent`.
- Salvaging a component can destroy the item, so it is re-read from its slot after each salvage. `SalvageComponents($buyKit, $dryRun)` has a dry-run mode that only logs what would be salvaged (`;SalvageComponents(False, True)` left in the test suite).
- In town the routine runs when `HasComponentsToSalvage()`; mid-run it runs without buying kits.
- The previously disabled GUI checkbox is now active with a tooltip; default farm and loot configurations, JSON read/write and README updated.

## Test plan

- [ ] Tick *Salvage into components*, run a farm dropping modded items, check that only items failing the keep rules are stripped and that the components reach storage
- [ ] Start without an expert kit in town and check one is bought; check nothing is bought mid-run
- [ ] Run `SalvageComponents(False, True)` from the test suite and review the dry-run log

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ To use it:
 ## Features
 - Unified interface shared by all bots 🖥️
 - Shared inventory, loot, farm, and title tracking 🎯
-- Configurable item handling (pickup, identify, salvage, sell, buy, store) 📦
+- Configurable item handling (pickup, identify, salvage, salvage into components, sell, buy, store) 📦
 - Farm UI with build, equipment, and contextual information 🛡️
 - Modular, plug-and-play bot system 🔌
 
--- a/lib/Utils-Storage.au3
+++ b/lib/Utils-Storage.au3
@@ -45,12 +45,13 @@ Func InventoryManagementBeforeRun()
 	; 2-Sort items
 	; 3-Identify items
 	; 4-Collect data
-	; 5-Salvage
-	; 6-Sell materials
-	; 7-Sell items
-	; 8-Balance character gold level
-	; 9-Buy ectoplasm/obsidian with surplus
-	; 10-Store items
+	; 5-Salvage valuable components out of items not worth keeping
+	; 6-Salvage
+	; 7-Sell materials
+	; 8-Sell items
+	; 9-Balance character gold level
+	; 10-Buy ectoplasm/obsidian with surplus
+	; 11-Store items
 	If $cache['Store items.Unidentified gold items'] And HasGoldUnidentifiedItems() Then
 		If GetMapType() <> $ID_OUTPOST Then TravelToOutpost($trade_town, $district_name)
 		StoreItemsInXunlaiStorage(IsUnidentifiedGoldItem)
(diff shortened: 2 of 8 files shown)


caustic-kronos:
Hello !
I wish we could add this, but last I tried adding it, it crashed the session (sends you back to character screen).
I pulled your PR and tried again, but I get the same result: it does salvage one mod, but then the session crashes.
Don't you have the same issue ?

lib/Utils-Storage.au3
```diff
@@ -541,6 +563,11 @@ Func DefaultShouldStoreItem($item)
 		Local $rarityName = $RARITY_NAMES_FROM_IDS[$rarity]
 		;Return ContainsValuableUpgrades($item)
 		Return $cache['Store items.Armor salvageables.' & $rarityName]
+	; ----------------------------------- Upgrade components -----------------------------------
+	ElseIf IsUpgradeComponent($item) Then
+		Local $shouldStore = $cache['Store items.Upgrade components']
+		; Loot configurations saved before this option existed have no value for it: store components when the bot salvages them
+		Return $shouldStore == Null ? $run_options_cache['run.salvage_into_components'] : $shouldStore
```
caustic-kronos, line 570: We can do without the backward compatibility - would be too much of a pain to keep configuration backward compatible.
```suggestion
		Return $cache['Store items.Upgrade components']
```

README.md
```diff
@@ -29,7 +29,7 @@ To use it:
 ## Features
 - Unified interface shared by all bots 🖥️
 - Shared inventory, loot, farm, and title tracking 🎯
-- Configurable item handling (pickup, identify, salvage, sell, buy, store) 📦
+- Configurable item handling (pickup, identify, salvage, salvage into components, sell, buy, store) 📦
```
caustic-kronos, line 32:
```suggestion
- Configurable item handling (pickup, identify, salvage components, salvage materials, sell, buy, store) 📦
```

lib/Utils-Storage.au3
```diff
@@ -1364,6 +1418,88 @@ Func SalvageItem($item, $salvageKit, $salvageType = 'Materials')
...
+;~ Salvaging a component can destroy the item, so it is re-read from its slot after each salvage
+;~ With $dryRun, only logs what would be salvaged
+Func SalvageComponents($buyKit = True, $dryRun = False)
+	; A weapon holds at most a prefix, a suffix and an inscription
+	Local Static $maxComponentsPerItem = 3
+	Info('Salvaging components out of items')
+	Local $kit = Null
+	Local $uses = 0
+	If Not $dryRun Then
+		$kit = GetExpertSalvageKit($buyKit)
```
caustic-kronos, line 1434: We should support infinite salvage kit - you can look into this great PR from Wellwisher to do the same with infinite salvage: https://github.com/caustic-kronos/BotsHub/pull/226

caustic-kronos (changes requested):
Hello, great PR !
There is not much to change, though I'm sure there are many ways we could improve performances here.
Reducing the _ArrayAdd, reusing some maps, removing others, reducing looping over structure, etc ... But honestly, out of scope, and it's good as is.

Unfortunately, as I wrote earlier, it can't be merged as long as it causes session crash/desync/character delogging. If you find a solution for it, then this PR will go through.

